PR #9, opened 2022-12-14
Rename our core stakeholders to "participants" (from "users")

fulldecent:
# Motivation

Many software projects name their customers as "users" and it seems this is from inertia rather than thoughtful consideration. I hope we can be more intentional with how we refer to the people that are using AT Protocol and this issue is for people also interested in the cause. I couldn't find the mission statement for the AT Protocol, but I assume it's something like "Social media must be resilient to corporate and government control, and empower the agency of the original author" ([inspiration](https://www.getrevue.co/profile/jackjack/issues/a-native-internet-protocol-for-social-media-1503112?via=twitter-card&client=DesktopWeb&element=issue-card#)). If so, this issue is in project scope whereas it helps further respect the agency of the original author.

The basic problems with the word choice "user" are:

1. **It is not descriptive**—*user* means lots of things, even in the context of a software project, the protocol producers and consumers, the humans behind it, robots, the server and other servers, the cron jobs, every part of the software itself can all be called users.
2. **Not human enough**—typically with software, we are imagining ways that a computer can serve a human, let's use more human words to describe the customers then. Are they publishers or readers? Let's call them that. Is this a community? Let's find some inspiration from that.
3. **"User" subjugates the customer's interests below the product's interests**—I have recently come to appreciate why you should call Bob as "looking for a house" or "displaced" or some of the other words people use today, rather than "a homeless person". Because Bob is not defined by his not having a home, that is a poor description for a person. Also, when "a homeless person" moves into a home they die (literally *and* figuratively, because the subject is figurative). Instead, let's choose names/words from the customer's perspective in a way that respects their interests.

# Background

Many people/agents will be using AT Protocol, including:

1. people/entities that publish content
1. people/entities that consume content
1. software agents that work on the behalf of 1 and 2
1. servers that provide storage and execution of the protocol
1. transport agents that connect to the servers, but not necessarily initiated by the command of 1 and 2
1. other things we have not dreamed up yet

# Ideas instead of "users"

- Participants
- Members
- Publishers
- Syndicates

And if we need to separate producers and consumers of the protocol, we can refer to the humans with agency as "publishers and readers" eschewing a single word.

---

This PR changes the word for the top file, and with permission/invitation I'm committing to update all the documentation to match.

### DIFF
--- a/content/guides/overview.md
+++ b/content/guides/overview.md
@@ -1,6 +1,7 @@
 ---
 title: Protocol Overview
 summary: An introduction to the AT Protocol.
+
 ---
 
 # Protocol Overview
@@ -9,13 +10,13 @@ The **Authenticated Transfer Protocol**, aka **ATP**, is a protocol for large-sc
 
 ## Identity
 
-Users are identified by domain names in AT Protocol. These domains map to cryptographic URLs which secure the user's account and its data.
+Participants are identified by domain names in AT Protocol. These domains map to cryptographic URLs which secure the participant's account and its data.
 
 ![Identities](/img/identities.jpg)
 
 ## Data repositories
 
-User data is exchanged in [signed data repositories](/guides/data-repos). These repositories are collections of records which include posts, comments, likes, follows, media blobs, etc.
+Participant data is exchanged in [signed data repositories](/guides/data-repos). These repositories are collections of records which include posts, comments, likes, follows, media blobs, etc.
 
 ![Data repos](/img/data-repos.jpg)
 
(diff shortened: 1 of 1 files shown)


harshsingh32 (approved):

ianklatzco:
The notion of "user" has strong cultural cachet (e.g. original TRON).

i.e., there's an extra semantic meaning in English with "user", that isn't necessarily present in other words.

A "user" is something special -- there's a hard-to-define semantic coordinate that's being pointed at when the word "user" is used.

I wonder if bsky could embed such a meaning into another word.

bnewbold:
This has gotten out of sync with the repo. Please open an issue to discuss first if you want to continue the conversation.

PR #11, opened 2023-01-05
Fix a few doc typos

Jacob2161:
Spotted a small number of typos while working through the docs and figured I'd fix 'em while I was looking.


pfrazee:
Nice, good catches

PR #25, opened 2023-04-13
fix typo on specs/did-plc page

vvakame:
change "... contains a reference the the immediately preceding ..." to "... contains a reference to the immediately preceding ..."


pfrazee:
Thanks!

PR #27, opened 2023-04-15
Added 404 and 500 error pages.

emrecoban:
I've added custom error pages for ATProtocol instead of users encountering the default error pages of Next.js.

Here's a sample:
404-page

cc: @arcalinea @pfrazee


bnewbold:

PR #29, opened 2023-04-21
Update Lexicon specs page

jiftechnify:
I found that the document for Lexicon specs is outdated, so updated it.  

I also updated the example of tokens to use a real-world schema.


bnewbold:
Thanks for this contribution! We ended up re-writing the lexicon language specs page with a different structure.

PR #36, opened 2023-04-24
Add visible :focus to Links on the /docs page for tab/shift+tab

ttrreevvoorr:
This PR adds visible focus to the Links on the /docs page. This resulted in a restructure of some of the JSX, and some additional tailwind class name additions.

While this doesn't meet all of WCAG's requirements for focus appearance, I think it's a step in the right direction. Source: https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance-minimum.html

The result appears like so in Firefox
ezgif-2-4028d355cc


ttrreevvoorr:
image

ttrreevvoorr:
I close by accident

ttrreevvoorr:
I'm closing this because someone implemented something like this already.

PR #38, opened 2023-04-24
Added alt attribute to hero image.

w3tsa:
img elements must have an alt prop, either with meaningful text, or an empty string for decorative images. eslint[jsx-a11y/alt-text](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/tree/HEAD/docs/rules/alt-text.md)


emilyliu7321:
this was fixed in #78, thanks for flagging!

PR #42, opened 2023-04-26
Add Lexicons for labeling embeds and Blocks

thinkverse:
Here's another Lexicon update @pfrazee. 👍


thinkverse:
PR has been updated to include the Lexicon for Blocks @pfrazee

pfrazee:
Thanks!

PR #49, opened 2023-05-03
Improves website accessibility

kokamkarsahil:
This PR does the following changes:
1. 061ee8058f85ee0b9b1def023ca8fb613af0db2e Adds the missing `lang=en` attribute to html tag.
~~f1f68f3e0f255260aa65c5ca511fe39eb3a238dc Defines explicit sizes of image. (Doesn't do any changes to img aspect ratio)~~
3. b67e1b57656747e1aa3940fde0f5574aec801f98 Adds meta description tag

### DIFF
--- a/components/meta.js
+++ b/components/meta.js
@@ -7,6 +7,7 @@ export default function Meta({ title, description, image }) {
   return (
     <Head>
       <title>{title}</title>
+      <meta name="description" content={description} />
       <meta property="og:title" content={title} />
       <meta property="og:type" content="article" />
       <meta
--- a/next.config.js
+++ b/next.config.js
@@ -5,6 +5,10 @@ const nextConfig = {
   images: {
     unoptimized: true,
   },
+  i18n: {
+    locales: ['en'],
+    defaultLocale: 'en',
+  },
   async redirects() {
     return [
       {
@@ -37,5 +41,6 @@ const nextConfig = {
     ]
   },
 }
+
 
 module.exports = nextConfig


components/header.js
```diff
@@ -58,7 +58,7 @@ export default function Header() {
           <Link href="/">
             <a tabIndex="0">
               <span className="sr-only">ATP</span>
-              <img className="h-10 w-auto" src="/logo.jpg" alt="" />
+              <img className="h-10 w-auto" src="/logo.jpg" alt="" width="288" height="96" />
```
bnewbold, line 61: setting image sizes is not something I have encountered before when doing web a11y work. this seems not-idiomatic, should be done with CSS if needed, and could mess with layout if we update the image files.

maybe i'm wrong! can you describe your motivation here?

bnewbold:
left a comment about img tag height/width, other change LGTM

components/header.js
kokamkarsahil: > setting image sizes is not something I have encountered before when doing web a11y work. this seems not-idiomatic, should be done with CSS if needed, and could mess with layout if we update the image files.

Yes, you are right, it isn't really an a11y thing. And about img changes, you're right about it. When the image is changed, the dimensions attribute needs to be updated.

> can you describe your motivation here?

It can greatly assist the web magnifiers and prevent layout changes on slow connections.

If this change is not required, please let me know so that I can revert this commit.

Thank you!

bnewbold: Yeah, please pull the image size changes out (revert the commit or rebase and force-push the branch), and this PR will be easier to merge.

Thanks!

kokamkarsahil: > Yeah, please pull the image size changes out (revert the commit or rebase and force-push the branch), and this PR will be easier to merge.

Sure, it's done.

Thanks for the review!

bnewbold (approved):

kokamkarsahil:
Sorry if this PR caused any issue, it looks the build failed for this PR.

But I am not able to reproduce it with the same commit 3e0691f from the main branch on the build step.

Please let me know if the build step causes any issue in the logs.

Thank you and really sorry again for any trouble!

emilyliu7321:
no problem and thanks for flagging, i'm going to temporarily revert these changes in the PR to isolate the build issue 🙏

PR #65, opened 2023-05-11
Add app.bsky.feed.getFeedSkeleton to lexicons page

emilyliu7321:


dholms (approved):
Looks good!

If there's anywhere to stick a short "subject to change", might be good. Otherwise probably fine

PR #74, opened 2023-05-14
community page: remove Bridgy Fed, add arroba

snarfed:
...since I extracted the ATP PDS code out of Bridgy Fed into its own library.

thanks for open sourcing the site so we can all file PRs!


emilyliu7321 (approved):
thank you!

PR #79, opened 2023-05-15
feat: add Get Alt Text bot to Community Projects

hs4man21:
I noticed some folks are using DIDs for the hyperlinks. I also noticed that many are using bsky.app instead of staging.bsky.app (though the most recent bot addition used staging, so I wasn't sure which is best right now).

Is there an official preference? I'm happy to adjust as needed!


emilyliu7321 (approved):
this format works, thank you!

PR #93, opened 2023-05-24
specs: add 'handle' spec doc

bnewbold:
This hasn't been formally reviewed by somebody else, but the handle identifier is pretty stable and this matches what the typescript `atproto` implementation does very closely. This document could probably also include docs on handle *resolution* process in addition to the handle *identifier*.

Mostly tossing this PR up to see how much work it is to go from an internal spec doc to basic public web doc. Answer, with notion markdown export, is basically trivial to get this far. Will want to do additional copy-editing and formatting, but there is also the option to just ship some of these fast to get the details out there, and polish presentation later.

### DIFF
--- a/content/specs/handle.md
+++ b/content/specs/handle.md
@@ -98,13 +98,13 @@ Handles may be prefixed with the “at” symbol (like `@jay.bsky.team`) in user
 
 Internationalized Domain Names (”IDN”, or “punycode”) are not directly relevant to the low-level handle syntax. In their encoded form, IDNs are already valid hostnames, and thus valid handles. Such handles must be stored and transmitted in encoded ASCII form. Handles that “look like” IDNs, but do not parse as valid IDNs, are valid handles, just as they are valid hostnames. Applications may, optionally, parse and display IDN handles as Unicode.
 
-Handles are not case-sensitive, which means they can be safely normalized from user input to lower-case (ASCII) form. Only normalized (lower-case) handles should be stored in records or used in outbound API calls. Applications should not preserve user-provided case information and attempt to display handles in anything other than lower-case. For example, the handle input string `[BlueskyWeb.xyz](http://www.BlueskyWeb.xyz)`  should be normalized, stored, and displayed as `blueskyweb.xyz`.
+Handles are not case-sensitive, which means they can be safely normalized from user input to lower-case (ASCII) form. Only normalized (lower-case) handles should be stored in records or used in outbound API calls. Applications should not preserve user-provided case information and attempt to display handles in anything other than lower-case. For example, the handle input string `BlueskyWeb.xyz`  should be normalized, stored, and displayed as `blueskyweb.xyz`.
 
 Very long handles are known to present user interface challenges, but they are allowed in the protocol, and application developers are expected to support them.
 
-Handles which look similar to a well-known domain present security and impersonation challenges. For example, handles like `[paypa1.com](http://paypa1.com/`paypal.com`)` or `[paypal.cc](http://paypal.cc)` being confused for `paypal.com`. Very long handles can result in similar issues when truncated at the start or end (`paypal.com…`).
+Handles which look similar to a well-known domain present security and impersonation challenges. For example, handles like `paypa1.com` or `paypal.cc` being confused for `paypal.com`. Very long handles can result in similar issues when truncated at the start or end (`paypal.com…`).
 
-Handles should generally not be truncated to local context. For example, the handle `@jay.[bsky.social](http://bsky.social)` should not be displayed as `@jay`, even in the local context of a `bsky.social` service.
+Handles should generally not be truncated to local context. For example, the handle `@jay.bsky.social` should not be displayed as `@jay`, even in the local context of a `bsky.social` service.
 
 Providers of handle “namespaces” (eg, as subdomains on a registered domain) may impose any additional limits on handles that they wish. It is recommended to constrain the allowed segment length to something reasonable, and to reserve a common set of segment strings like `www`, `admin`, `mail`, etc. There are multiple public lists of “commonly disallowed usernames” that can be used as a starting point.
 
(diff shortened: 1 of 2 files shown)


bnewbold:
Preview URL: https://bnewbold-spec-handle.atproto-website.pages.dev/specs/handle

content/specs/handle.md
```diff
@@ -0,0 +1,119 @@
...
+blah.arpa
+```
+
+### Usage and Implementation Guidelines
+
+Handles may be prefixed with the “at” symbol (like `@jay.bsky.team`) in user interfaces, but this is not a valid syntax for a handle in records, APIs, and other back-end contexts.
+
+Internationalized Domain Names (”IDN”, or “punycode”) are not directly relevant to the low-level handle syntax. In their encoded form, IDNs are already valid hostnames, and thus valid handles. Such handles must be stored and transmitted in encoded ASCII form. Handles that “look like” IDNs, but do not parse as valid IDNs, are valid handles, just as they are valid hostnames. Applications may, optionally, parse and display IDN handles as Unicode.
+
+Handles are not case-sensitive, which means they can be safely normalized from user input to lower-case (ASCII) form. Only normalized (lower-case) handles should be stored in records or used in outbound API calls. Applications should not preserve user-provided case information and attempt to display handles in anything other than lower-case. For example, the handle input string `[BlueskyWeb.xyz](http://www.BlueskyWeb.xyz)`  should be normalized, stored, and displayed as `blueskyweb.xyz`.
```
ianklatzco, line 101: Should this be a markdown link or just `BlueskyWeb.xyz` ?

ianklatzco:
Added links to RFCs.

patch: paste this into a file and then `git apply FILENAME`

```
diff --git a/content/specs/handle.md b/content/specs/handle.md
index 451ed13..fccfc2f 100644
--- a/content/specs/handle.md
+++ b/content/specs/handle.md
@@ -7,7 +7,7 @@ summary: A specification for human-friendly account identifiers.

 DIDs are the long-term persistent identifiers for accounts in atproto, but they can be opaque and unfriendly for human use. Handles are a less-permanent identifier for accounts. The mechanism for verifying the link between an account handle and an account DID relies on DNS, and possibly connections to a network host, so every handle must be a valid network hostname. *Almost* every valid “hostname” is also a valid handle, though there are a small number of exceptions.

-The definition “hostnames” (as a subset of all possible “DNS names”) has evolved over time and across several RFCs. Some relevant documents are RFC-1035, RFC-3696 section 2, and RFC-3986 section 3.
+The definition “hostnames” (as a subset of all possible “DNS names”) has evolved over time and across several RFCs. Some relevant documents are [RFC-1035](https://datatracker.ietf.org/doc/html/rfc1035), [RFC-3696 section 2](https://datatracker.ietf.org/doc/html/rfc3696#autoid-2), and [RFC-3986 section 3](https://datatracker.ietf.org/doc/html/rfc3986#autoid-17).

 ### Handle Syntax
```

content/specs/handle.md
ianklatzco: Also applies to links further down (paypal, jay.bsky.social)

bnewbold: thanks!

content/specs/handle.md
```diff
@@ -0,0 +1,119 @@
...
+- the overall handle is split in to multiple segments (referred to as “labels” in standards documents), separated by ASCII periods (`.`)
+- no proceeding or trailing ASCII periods are allowed, and there must be at least two segments. that is, ”bare” top-level domains are not allowed as handles, even if valid “hostnames” and “DNS names”. and “tailing dot” syntax for DNS names is not allowed for handles
+- each segment must have at least 1 and at most 63 characters (not including the periods). the allowed characters are ASCII letters (`a-z`), digits (`0-9`), and hyphens (`-`)
+- segments can not start or end with a hyphen.
+- the last segment (the “top level domain”) can not start with a numeric digit
+- handles are not case-sensitive, and should be normalized to lower-case (that is, normalize ASCII `A-Z` to `a-z`)
+
+To be explicit (the above rules already specify this), no whitespace, null bytes, joining characters, or other ASCII control characters are allowed in the handle, including as prefix/suffix.
+
+Modern “hostnames” (and thus handles) allow ASCII digits in most positions, with the exception that the last segment (top-level domain, TLD) can not start with a digit.
```
emilyliu7321, line 28:
```suggestion
Modern “hostnames” (and thus handles) allow ASCII digits in most positions, with the exception that the last segment (top-level domain, TLD) cannot start with a digit.
```

content/specs/handle.md
```diff
@@ -0,0 +1,119 @@
...
+
+Handles should generally not be truncated to local context. For example, the handle `@jay.bsky.social` should not be displayed as `@jay`, even in the local context of a `bsky.social` service.
+
+Providers of handle “namespaces” (eg, as subdomains on a registered domain) may impose any additional limits on handles that they wish. It is recommended to constrain the allowed segment length to something reasonable, and to reserve a common set of segment strings like `www`, `admin`, `mail`, etc. There are multiple public lists of “commonly disallowed usernames” that can be used as a starting point.
+
+From a practical standpoint, handles should be limited to at most 244 characters, fewer than the 253 allowed for DNS names. This is because DNS verification works with the prefix `_atproto.`, which adds 9 characters, and that overall name needs to be valid.
+
+Handle hostnames are expected to be mainstream DNS domain names, registered through the mainstream DNS name system. Handles with non-standard TLDs, or using non-standard naming systems, will fail to interoperate with other network services and protocol implementations in the atproto ecosystem.
+
+### Possible Future Changes
```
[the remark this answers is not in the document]
bnewbold: That is a real concern. I think we can't realistically retain VariableCase through the entire system, and having inconsistent display of the same identifier raises it's own accessibility and usability issues.

I added to the relevant "usage guidelines" paragraph:

> Long all-lowercase handles can be a readability and accessibility challenge. Sub-domain separation (periods), hyphenation, or use of "display names" in application protocols can all help.

This is kind of a band-aid, but I think it is realistically how people real with this with actual domain names: break.things.up.com or break-things-up.com instead of breakthingsup.com or BreakThingsUp.com

emilyliu7321 (approved):
lgtm, thank you!

PR #102, opened 2023-06-02
specs/did: replace specs/did-plc with generic doc

bnewbold:
At a high level, moved did:plc stuff to the `did-method-plc` repo, via a redirect of the old spec URL, and new links. Then, add a new DID spec document that describes our use of DIDs a bit more generically (though still with mention of the particular methods).

A missing piece here that I would add later in a second pass is more details about what fields to parse out of DID documents for use with atproto.

Not much new here, mostly just specifying existing behavior. The did:web lockdown stuff was implemented upstream in atproto (Typescript).


content/specs/did.md
```diff
@@ -0,0 +1,93 @@
+---
+title: DID
+summary: Persistent decentralized identifiers (as used in atproto)
+---
+
+# DID (in atproto)
+
+The AT Protocol uses [Decentralized IDentifiers](https://en.wikipedia.org/wiki/Decentralized_identifier) (DIDs) as persistent, long-term account identifiers. DID is a W3C standard, with many standardized and proposed "DID method" implementations.
```
emilyliu7321, line 8:
```suggestion
The AT Protocol uses [Decentralized Identifiers](https://en.wikipedia.org/wiki/Decentralized_identifier) (DIDs) as persistent, long-term account identifiers. DID is a W3C standard, with many standardized and proposed DID method implementations.
```

content/specs/did.md
```diff
@@ -0,0 +1,93 @@
...
+- `did:plc`, which is a novel DID method developed by Bluesky. See the [did-method-plc](https://github.com/bluesky-social/did-method-plc) github repository for details. Even though this method is referred to as "placeholder", it will be supported indefinitely.
+
+In the future, a *small* number of additional methods may be supported. It is not the intention to support "all" or even "many" DID methods, even with the existence of "universal resolver" software.
```
emilyliu7321, line 18:
```suggestion
Currently, atproto supports two DID methods:

- `did:web`, which is a W3C standard based on HTTPS (and DNS). The identifier section is a hostname. This method is supported in atproto to provide an independent alternative to `did:plc`. The method is inherently tied to the domain name used, and does not provide a mechanism for migration or recovering from loss of control of the domain name. In the context of atproto, only hostname-level `did:web` DIDs are supported: path-based DIDs are not supported. The same restrictions on top-level domains that apply to handles (eg, no `.arpa`) also apply to `did:web` domains. The special `[localhost](http://localhost)` hostname is allowed, but only in testing and development environments. Port numbers (with separating colon hex-encoded) are only allowed for `localhost`, and only in testing and development.
- `did:plc`, which is a novel DID method developed by Bluesky. See the [did-method-plc](https://github.com/bluesky-social/did-method-plc) GitHub repository for details. Even though this method is referred to as "placeholder", it will be supported indefinitely.

In the future, a small number of additional methods may be supported. It is not the intention to support all or even many DID methods, even with the existence of universal resolver software.
```

content/specs/did.md
```diff
@@ -0,0 +1,93 @@
+---
+title: DID
+summary: Persistent decentralized identifiers (as used in atproto)
+---
+
+# DID (in atproto)
```
emilyliu7321, line 6:
```suggestion
# AT Protocol DIDs
```

content/specs/did.md
```diff
@@ -0,0 +1,93 @@
...
+DID identifiers do not generally have a maximum length restriction. In the context of atproto there is an initial hard limit of 2 KBytes.
+
+In the context of atproto, implementations do not need to validate percent encoding. The percent symbol is allowed in DID "identifier" segments, but the identifier should not end in a percent symbol. A DID containing invalid percent encoding *should* fail any attempt at registration, resolution, etc.
+
+A reasonable starting-point regex for DIDs in the context of atproto is:
+
+```
+// NOTE: does not constrain overall length
+/^did:[a-z]+:[a-zA-Z0-9._:%-]*[a-zA-Z0-9._-]$/
+```
```
emilyliu7321, line 45:
```suggestion
### AT Protocol DID Syntax

Lexicon string type: `did`

The DID Core specification constraints on DID identifier syntax, regardless of the method used. A summary of those syntax constraints, which may be used to validate DID generically in atproto are:

- The entire URI is made up of a subset of ASCII, containing letters (`A-Z`, `a-z`), digits (`0-9`), period, underscore, colon, percent sign, or hyphen (`._:%-`)
- The URI is case-sensitive
- The URI starts with lowercase `did:`
- The method segment is one or more lowercase letters (`a-z`), followed by `:`
- The remainder of the URI (the identifier) may contain any of the above-allowed ASCII characters, except for percent-sign (`%`)
- The URI (and thus the remaining identifier) may not end in `:`.
- Percent-sign (`%`) is used for "percent encoding" in the identifier section, and must always be followed by two hex characters
- Query (`?`) and fragment (`#`) sections are allowed in DID URIs, but not in DID identifiers. In the context of atproto, the query and fragment parts are not allowed.

DID identifiers do not generally have a maximum length restriction, but in the context of atproto, there is an initial hard limit of 2 KB.

In the context of atproto, implementations do not need to validate percent encoding. The percent symbol is allowed in DID identifier segments, but the identifier should not end in a percent symbol. A DID containing invalid percent encoding *should* fail any attempt at registration, resolution, etc.

A reasonable starting-point regex for DIDs in the context of atproto is:

```
// NOTE: does not constrain overall length
/^did:[a-z]+:[a-zA-Z0-9._:%-]*[a-zA-Z0-9._-]$/
```
```

emilyliu7321 (approved):

PR #112, opened 2023-06-10
specs/repository: correction about skipped intermediate nodes

bnewbold:
Quick fix, thanks @DavidBuchanan314 for catching the inconsistency between the written spec and the typescript (and golang) implementation.


dholms (approved):
👍

PR #114, opened 2023-06-13
specs/event-stream: initial specification

bnewbold:
I think @dholms may already have given this a read-over. Almost no changes vs the notion version from last week.

The one major change I could see making here is using "Subscription" instead of "Event Stream".

### DIFF
--- a/content/specs/event-stream.md
+++ b/content/specs/event-stream.md
@@ -1,6 +1,6 @@
 ---
 title: Event Stream
-summary: Wire protocol for streaming Lexicon objects over a network connection
+summary: Network wire protocol for subscribing to a stream of Lexicon objects
 wip: true
 ---
 
@@ -10,11 +10,13 @@ In addition to regular [HTTP API](/specs/xrpc) endpoints, atproto supports conti
 
 The Lexicon type for streams is `subscription`. The schema includes an identifier (`id`) for the endpoint, a `message` schema (usually a union, allowing multiple message types), and a list of error types (`errors`).
 
-Clients subscribe to a specific stream by initiating a connection at the indicated endpoint. Streams are currently one-way, with messages only flowing from the server to the client. Clients may provide query parameters to configure the stream when opening the connection.
+Clients subscribe to a specific stream by initiating a connection at the indicated endpoint. Streams are currently one-way, with messages flowing from the server to the client. Clients may provide query parameters to configure the stream when opening the connection.
 
 A **backfill window** mechanism allows clients to catch up with stream messages they may have missed. At a high level, this works by assigning monotonically increasing sequence numbers to stream events, and allowing clients to specify an initial sequence number when initiating a connection. The intent of this mechanism is to ensure reliable delivery of events following disruptions during a reasonable time window (eg, hours or days). It is not to enable clients to roll all the way back to the beginning of the stream.
 
-All of the streaming APIs we have now or are considering use backfill. However, a backfill mechanism (and even cursors, which we define below) is not _required_ for streams. For example, there could be a streaming API that has no concerns for dropping or redelivering events.
+All of the initial subscription Lexicons in the `com.atproto` namespace use the backfill mechanism. However, a backfill mechanism (and even cursors, which we define below) is not _required_ for streams. Subscription endpoints which do not require reliable delivery do not need to implement a backfill mechanism or use sequence numbers.
+
+The initial subscription endpoints are also public and do not require authentication or prior permission to subscribe (though resource limits may be imposed on client). But subscription endpoints may require authentication at connection time, using the existing HTTP API (XRPC) authentication methods.
 
 
 ## Streaming Wire Protocol (v0)
@@ -88,9 +90,7 @@ The scope for sequence numbers is the combination of service provider (hostname)
 
 Services should ensure that sequence numbers are not re-used, usually by committing events (with sequence number) to robust persistent storage before transmitting them over streams.
 
-If a server deletes all their sequenced events or their service, or suffers a hard crash and loses some events, the client needs to be able to recover a connection with them. In this case, the server will send back a `FutureCursor` error to the client. The client can then choose what to do. 
-
-One option for the client is restart at the earliest backfill time. We suggest that clients treat out-of-order or duplicate sequence numbers as an error, not process the message, and drop the connection. This includes across re-connection attempts. Clients should not reset sequence state without human operator intervention.
+In some catastrophic failure modes, or large changes in infrastructure, it is possible that a server would both data from the backfill window, and need to reset the sequence number back to `1`. In this case, if a client re-connects with a higher number, the server would send back a `FutureCursor` error to the client. The client needs to decide what strategy to follow in these scenarios. We suggest that clients treat out-of-order or duplicate sequence numbers as an error, not process the message, and drop the connection. Most clients should not reset sequence state without human operator intervention, though this may be a reasonable behavior for some ephemeral clients not requiring reliable delivery of every event in the stream.
 
 ## Usage and Implementation Guidelines
 
--- a/content/specs/event-stream.md
+++ b/content/specs/event-stream.md
@@ -14,13 +14,14 @@ Clients subscribe to a specific stream by initiating a connection at the indicat
 
 A **backfill window** mechanism allows clients to catch up with stream messages they may have missed. At a high level, this works by assigning monotonically increasing sequence numbers to stream events, and allowing clients to specify an initial sequence number when initiating a connection. The intent of this mechanism is to ensure reliable delivery of events following disruptions during a reasonable time window (eg, hours or days). It is not to enable clients to roll all the way back to the beginning of the stream.
 
-Streams are open by default. No authentication or prior permission is required to subscribe, though resource limits may be imposed at large scales.
+All of the streaming APIs we have now or are considering use backfill. However, a backfill mechanism (and even cursors, which we define below) is not _required_ for streams. For example, there could be a streaming API that has no concerns for dropping or redelivering events.
+
 
 ## Streaming Wire Protocol (v0)
 
 To summarize, messages are encoded as DAG-CBOR and sent over a binary WebSocket. Clients connect to to a specific HTTP endpoint, with query parameters, then upgrade to WebSocket. Every WebSocket frame contains two DAG-CBOR objects, with bytes concatenated together: a header (indicating message type), and the actual message.
 
-The WebSockets "living standard" is currently maintained by [WHATWG](https://en.wikipedia.org/wiki/WHATWG), and can be found in full at: [https://websockets.spec.whatwg.org/](https://websockets.spec.whatwg.org/)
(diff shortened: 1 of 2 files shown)


content/specs/event-stream.md
```diff
@@ -0,0 +1,133 @@
...
+
+The Lexicon type for streams is `subscription`. The schema includes an identifier (`id`) for the endpoint, a `message` schema (usually a union, allowing multiple message types), and a list of error types (`errors`).
+
+Clients subscribe to a specific stream by initiating a connection at the indicated endpoint. Streams are currently one-way, with messages only flowing from the server to the client. Clients may provide query parameters to configure the stream when opening the connection.
+
+A **backfill window** mechanism allows clients to catch up with stream messages they may have missed. At a high level, this works by assigning monotonically increasing sequence numbers to stream events, and allowing clients to specify an initial sequence number when initiating a connection. The intent of this mechanism is to ensure reliable delivery of events following disruptions during a reasonable time window (eg, hours or days). It is not to enable clients to roll all the way back to the beginning of the stream.
```
dholms, line 15: A backfill mechanism (& even cursors) are not required for streams.

You could imagine a streaming api that has no concerns for dropping or redelivering events.

However, all of the streaming apis we have now - or are considering, as is the case with subscribeLabels -  do use backfill

content/specs/event-stream.md
```diff
@@ -0,0 +1,133 @@
...
+
+Streams are open by default. No authentication or prior permission is required to subscribe, though resource limits may be imposed at large scales.
```
dholms, line 17: again these are features of the particular subscriptions not streams in general

content/specs/event-stream.md
```diff
@@ -0,0 +1,133 @@
...
+- `cursor` is older than roll-back window: the first message in stream is an info indicating that `cursor` is too-old, then starts at the oldest available `seq` and sends the entire roll-back window, then continues with current stream
+- `cursor` is `0`: server will start at the oldest available `seq`, send the entire roll-back window, then continue with current stream
+
+The scope for sequence numbers is the combination of service provider (hostname) and endpoint (NSID). This roughly corresponds to the `wss://` URL used for connections. That is, sequence numbers may or may not be unique across different stream endpoints on the same service.
+
+Services should ensure that sequence numbers are not re-used, usually by committing events (with sequence number) to robust persistent storage before transmitting them over streams.
+
+Clients should treat out-of-order or duplicate sequence numbers as an error, not process the message, and drop the connection. This includes across re-connection attempts. Clients should not reset sequence state without human operator intervention.
```
dholms, line 90: I think there's some subtlety here - if a server deletes all their sequenced events, or their service, or suffers a hard crash & loses some events, the client needs to be able to recover a connection with them

In this case, the server will send back a `FutureCursor` error to the client. The client can then choose what to do, one option would be to restart at the earliest backfill time

dholms:
No strong opinion on subscription or event stream. Just that we use "subscription" as the lexicon type & in the method names so it seems like the more natural choice

content/specs/event-stream.md
emilyliu7321:
```suggestion
A **backfill window** mechanism allows clients to catch up with stream messages they may have missed. At a high level, this works by assigning monotonically increasing sequence numbers to stream events, and allowing clients to specify an initial sequence number when initiating a connection. The intent of this mechanism is to ensure reliable delivery of events following disruptions during a reasonable time window (eg, hours or days). It is not to enable clients to roll all the way back to the beginning of the stream.

All of the streaming APIs we have now or are considering use backfill. However, a backfill mechanism (and even cursors, which we define below) is not _required_ for streams. For example, there could be a streaming API that has no concerns for dropping or redelivering events.
```

content/specs/event-stream.md
emilyliu7321:
```suggestion
```

content/specs/event-stream.md
```diff
@@ -0,0 +1,133 @@
...
+
+Streams are open by default. No authentication or prior permission is required to subscribe, though resource limits may be imposed at large scales.
+
+## Streaming Wire Protocol (v0)
+
+To summarize, messages are encoded as DAG-CBOR and sent over a binary WebSocket. Clients connect to to a specific HTTP endpoint, with query parameters, then upgrade to WebSocket. Every WebSocket frame contains two DAG-CBOR objects, with bytes concatenated together: a header (indicating message type), and the actual message.
+
+The WebSockets "living standard" is currently maintained by [WHATWG](https://en.wikipedia.org/wiki/WHATWG), and can be found in full at: [https://websockets.spec.whatwg.org/](https://websockets.spec.whatwg.org/)
```
emilyliu7321, line 23:
```suggestion
The WebSockets "living standard" is currently maintained by [WHATWG](https://en.wikipedia.org/wiki/WHATWG), and can be found in full at [https://websockets.spec.whatwg.org/](https://websockets.spec.whatwg.org/).
```

content/specs/event-stream.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Connection
+
+Clients initialize stream subscriptions by opening an HTTP connection and upgrading to a WebSocket. HTTPS and "WebSocket Secure" (`wss://`) on the default port (443) should be used for all connections on the internet. HTTP, cleartext WebSocket (`ws://`), and non-standard ports should only be used for testing, development, and local connections (for example, behind a reverse proxy implementing SSL). From the client perspective, failure to upgrade connection to a WebSocket is an error.
+
+Query parameters may be provided in the initial HTTP request to configure the stream in an application-specific way, as specified in the endpoint's Lexicon schema.
+
+Errors are usually returned through the stream itself. Connection-time error are sent as the first message on the stream, and then the server drops the connection. But some errors can not be handled through the stream, and are returned as HTTP errors:
```
emilyliu7321, line 31:
```suggestion
Errors are usually returned through the stream itself. Connection-time errors are sent as the first message on the stream, and then the server drops the connection. But some errors can not be handled through the stream, and are returned as HTTP errors:
```

content/specs/event-stream.md
```diff
@@ -0,0 +1,133 @@
...
+- `405 Method Not Allowed`: Returned to client for non-GET HTTP requests to a stream endpoint.
+- `426 Upgrade Required`: Returned to client if if `Upgrade` header is not included in a request to a stream endpoint.
+- `429 Too Many Requests`: Frequently used for rate-limiting. Client may try again after a delay. Support for the `Retry-After` header is encouraged.
+- `500 Internal Server Error`: Client may try again after a delay
+- `501 Not Implemented`: Service does not implement WebSockets or streams, at least for this endpoint. Client should not try again.
+- `502 Bad Gateway`, `503 Service Unavailable`, `504 Gateway Timeout`: Client may try again after a delay
+
+Servers *should* return HTTP bodies as JSON with the standard XRPC error message schema for these status codes. But clients also need to be robust to unexpected response body formats. A common situation is receiving a default load-balancer or reverse-proxy error page during scheduled or unplanned downtime.
+
+Either the server or the client may decided to drop an open stream connection if there have been no messages for some time. It is also acceptable to leave connections open indefinitely.
```
emilyliu7321, line 42:
```suggestion
- `405 Method Not Allowed`: Returned to client for non-GET HTTP requests to a stream endpoint.
- `426 Upgrade Required`: Returned to client if `Upgrade` header is not included in a request to a stream endpoint.
- `429 Too Many Requests`: Frequently used for rate-limiting. Client may try again after a delay. Support for the `Retry-After` header is encouraged.
- `500 Internal Server Error`: Client may try again after a delay
- `501 Not Implemented`: Service does not implement WebSockets or streams, at least for this endpoint. Client should not try again.
- `502 Bad Gateway`, `503 Service Unavailable`, `504 Gateway Timeout`: Client may try again after a delay

Servers *should* return HTTP bodies as JSON with the standard XRPC error message schema for these status codes. But clients also need to be robust to unexpected response body formats. A common situation is receiving a default load-balancer or reverse-proxy error page during scheduled or unplanned downtime.

Either the server or the client may decided to drop an open stream connection if there have been no messages for some time. It is also acceptable to leave connections open indefinitely.
```

content/specs/event-stream.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Framing
+
+Each binary WebSocket frame contains two DAG-CBOR objects, concatenated. The first is a "header" and the second is the "payload".
```
emilyliu7321, line 46:
```suggestion
Each binary WebSocket frame contains two DAG-CBOR objects, concatenated. The first is a **header** and the second is the **payload.**
```

content/specs/event-stream.md
```diff
@@ -0,0 +1,133 @@
...
+
+The header DAG-CBOR object has the following fields:
+
+- `op` ("operation", integer, required): fixed values, indicating what this frame contains
+    - `1`: a regular message, with type indicated by `t`
+    - `-1`: an error message
+- `t` ("type", string, optional): required if `op` is `1`, indicating the Lexicon sub-type for this message, in short form. Does not include the full Lexicon identifier, just a fragment. Eg: `#commit`. Should not be included in header if `op` is `1`
```
emilyliu7321, line 53:
```suggestion
- `t` ("type", string, optional): required if `op` is `1`, indicating the Lexicon sub-type for this message, in short form. Does not include the full Lexicon identifier, just a fragment. Eg: `#commit`. Should not be included in header if `op` is `1`.
```

content/specs/event-stream.md
```diff
@@ -0,0 +1,133 @@
...
+Clients should ignore frames with headers that have unknown `op` or `t` values. Unknown fields in both headers and payloads should be ignored. Invalid framing or invalid DAG-CBOR encoding are hard errors, and the client should drop the entire connection instead of skipping the frame. Servers should ignore any frames received from the client, not treat them as errors.
+
+Error payloads all have the following fields:
+
+- `error` (string, required): the error type name, with no namespace or `#` prefix
+- `message` (string, optional): a description of the error
+
+Streams should be closed immediately following transmitting or receiving an error frame.
+
+Message payloads must always be objects. They should omit the `$type` field, as this information is already indicated in the header. There is no specific limit on the size of WebSocket frames in atproto, but they should be kept reasonably small. A couple megabytes is an informal guideline.
```
emilyliu7321, line 64:
```suggestion
Message payloads must always be objects. They should omit the `$type` field, as this information is already indicated in the header. There is no specific limit on the size of WebSocket frames in atproto, but they should be kept reasonably small (around a couple megabytes).
```

content/specs/event-stream.md
emilyliu7321:
```suggestion
If a server deletes all their sequenced events or their service, or suffers a hard crash and loses some events, the client needs to be able to recover a connection with them. In this case, the server will send back a `FutureCursor` error to the client. The client can then choose what to do. 

One option for the client is restart at the earliest backfill time. We suggest that clients treat out-of-order or duplicate sequence numbers as an error, not process the message, and drop the connection. This includes across re-connection attempts. Clients should not reset sequence state without human operator intervention.
```

bnewbold:
I re-worded a bit to use "subscription" terminology more, though the document over all is still call "event stream" and uses "streaming" in a bunch of places.

I could go either way on this and am open to iterating to completely change the name of the document in the future. I think it is clear that at least both terms are permissible and they are the same concept, and I don't think the confusion is urgent enough to block publishing for now.

PR #115, opened 2023-06-13
specs/data-model: initial specification

bnewbold:
Didn't change much since the internal notion version. Decided to move string formatting details over to the Lexicon doc, as those are more about validation against a specific Lexicon. Did add an explicit note about "nullable" and "false-y" values.

cc: @dholms @devinivy

### DIFF
--- a/content/specs/data-model.md
+++ b/content/specs/data-model.md
@@ -15,9 +15,9 @@ The schema definition language for atproto is [Lexicon](/specs/lexicon). The IPL
 
 In IPLD, distinct pieces of data are called **nodes,** and when encoded in binary (DAG-CBOR) result in a **block.** A node may have internal nested structure (maps or lists). Nodes may reference each other by string URLs or URIs, just like with regular JSON on the web. In IPLD, they can also reference each other strongly by hash, referred to in IPLD as a **link.** A set of linked nodes can form higher-level data structures like [Merkle Trees](https://en.wikipedia.org/wiki/Merkle_tree) or [Directed Acyclical Graphs (DAG)](https://en.wikipedia.org/wiki/Directed_acyclic_graph). Links can also refer to arbitrary binary data (blobs).
 
-Links are encoded as CIDs here. You can tell what is behind the data by looking at the codec of the CID (whether it is DAG-CBOR or arbitrary binary).
-
 Unlike URLs, hash references (links) do not encode a specific network location where the content can be found. The location and access mechanism must be inferred by protocol-level context. Hash references do have the property of being "self-certifying", meaning that returned data can be verified against the link hash. This makes it possible to redistribute content and trust copies even if coming from an untrusted party.
+
+Links are encoded as [IPFS Content Identifiers](https://docs.ipfs.tech/concepts/content-addressing/#identifier-formats) (CIDs), which have both binary and string representations. CIDs include a metadata code which indicates whether it links to a node (DAG-CBOR) or arbitrary binary data. Some additional constraints on the use of CIDs in atproto are described below.
 
 In atproto, object nodes often include a string field `$type` that specifies their Lexicon schema. Data is mostly self-describing and can be processed in schema-agnostic ways (including decoding and re-encoding), but can not be fully validated without the schema on-hand or known ahead of time.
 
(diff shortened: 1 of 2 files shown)


content/specs/data-model.md
```diff
@@ -0,0 +1,146 @@
...
+
+In IPLD, distinct pieces of data are called "nodes", and when encoded in binary (DAG-CBOR) result in a "block". A node may have internal nested structure (maps or lists). Nodes may reference each other by string URLs or URIs, just like with regular JSON on the web. In IPLD, they can also reference each other strongly by hash, referred to in IPLD as a "link". A set of linked nodes can form higher-level data structures like [Merkle Trees](https://en.wikipedia.org/wiki/Merkle_tree) or [Directed Acyclical Graphs (DAG)](https://en.wikipedia.org/wiki/Directed_acyclic_graph). Links can also refer to arbitrary binary data ("blobs"). Unlike URLs, hash references (links) do not encode a specific network location where the content can be found. The location and access mechanism must be inferred by protocol-level context. Hash references do have the property of being "self-certifying", meaning that returned data can be verified against the link hash. This makes it possible to redistribute content and trust copies even if coming from an untrusted party.
```
dholms, line 16: maybe worth a note about how "links" are encoded as CIDs here. and you can tell what is behind the data by looking at the codec of the cid - ie whether it is dag-cbor or arbitrary binary

content/specs/data-model.md
```diff
@@ -0,0 +1,146 @@
...
+There is also a deprecated legacy blob format, with some records in the wild still containing blob references in this format:
+
+- `cid` (string, required): a CID in *string* format, not *link* format
+- `mimeType` (string, required, not empty): same as `mimeType` above
+
+Note that the legacy format has no `$type` and can only be parsed for known Lexicons. Implementations should not throw errors when encountering the old format, but should never write them, and it is acceptable to only partially support them.
+
+## JSON Representation
+
+atproto uses it's own conventions for JSON, instead of using DAG-JSON directly. Normalizations like key sorting are also not required or enforced: only DAG-JSON is used as a byte-reproducible representation.
```
dholms, line 79: should we give a quick rationale for this?

also maybe point out that dag-json explicitly states that it's primarily useful for debugging & dev environments?

content/specs/data-model.md
```diff
@@ -0,0 +1,146 @@
...
+summary: Consistent data encoding for records and messages.
+---
+
+# Data Model
+
+Records and messages in atproto are stored, transmitted, encoded, and authenticated in a consistent way. The core "data model" is based on [Interplanetary Linked Data (IPLD)](https://ipld.io/docs/data-model/), a specification for hash-linked data structures from the IPFS ecosystem.
+
+When data needs to be authenticated (signed), referenced (linked by content hash), or stored efficiently, it is encoded in Concise Binary Object Representation (CBOR). CBOR is an IETF standard roughly based on JSON. IPLD specifies a normalized subset of CBOR called "DAG-CBOR", which is what atproto uses.
+
+IPLD also specifies an analogous set of conventions of JSON called "DAG-JSON", but atproto uses a different set of conventions when encoding JSON data.
```
emilyliu7321, line 12:
```suggestion
When data needs to be authenticated (signed), referenced (linked by content hash), or stored efficiently, it is encoded in Concise Binary Object Representation (CBOR). CBOR is an IETF standard roughly based on JSON. IPLD specifies a normalized subset of CBOR called **DAG-CBOR,** which is what atproto uses.

IPLD also specifies an analogous set of conventions of JSON called **DAG-JSON,** but atproto uses a different set of conventions when encoding JSON data.
```

content/specs/data-model.md
emilyliu7321:
```suggestion
In IPLD, distinct pieces of data are called **nodes,** and when encoded in binary (DAG-CBOR) result in a **block.** A node may have internal nested structure (maps or lists). Nodes may reference each other by string URLs or URIs, just like with regular JSON on the web. In IPLD, they can also reference each other strongly by hash, referred to in IPLD as a **link.** A set of linked nodes can form higher-level data structures like [Merkle Trees](https://en.wikipedia.org/wiki/Merkle_tree) or [Directed Acyclical Graphs (DAG)](https://en.wikipedia.org/wiki/Directed_acyclic_graph). Links can also refer to arbitrary binary data (blobs).

Links are encoded as CIDs here. You can tell what is behind the data by looking at the codec of the CID (whether it is DAG-CBOR or arbitrary binary).

Unlike URLs, hash references (links) do not encode a specific network location where the content can be found. The location and access mechanism must be inferred by protocol-level context. Hash references do have the property of being "self-certifying", meaning that returned data can be verified against the link hash. This makes it possible to redistribute content and trust copies even if coming from an untrusted party.
```

content/specs/data-model.md
emilyliu7321, line 79:
```suggestion
atproto uses its own conventions for JSON, instead of using DAG-JSON directly. Normalizations like key sorting are also not required or enforced: only DAG-JSON is used as a byte-reproducible representation.
```

content/specs/data-model.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Link and CID Formats
+
+The [IPFS CID specification](https://github.com/multiformats/cid) is very flexible. It supports a wide variety of hash types, a field indicating the "type" of content being linked to, and various string encoding options. These features are valuable to allow evolution over time, but to maximize interoperability among implementations, only a specific "blessed" set of CID types are allowed.
+
+The blessed formats for CIDs in atproto are:
+
+- CIDv1
+- multibase: binary serialization within DAG-CBOR `link` fields, and `base32` for string encoding
+- multicodec: `dag-cbor` (0x71) for links to data objects, and `raw` (0x55) for links to blobs
```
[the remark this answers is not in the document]
bnewbold: I'm hesitant to mention a specific hash (or folks might start using it early), but I clarified that we will probably add to the "blessed" CID hash types, specifically when referencing large media blobs. Both here and below in "future changes".

PR #116, opened 2023-06-15
specs: handle resolution and did document details

bnewbold:
Combined, a bunch of additional specs on how to work with identity in atproto.

### DIFF
--- a/content/specs/did.md
+++ b/content/specs/did.md
@@ -8,7 +8,7 @@ summary: Persistent decentralized identifiers (as used in atproto)
 The AT Protocol uses [Decentralized Identifiers](https://en.wikipedia.org/wiki/Decentralized_identifier) (DIDs) as persistent, long-term account identifiers. DID is a W3C standard, with many standardized and proposed DID method implementations.
 
 
-### Blessed DID Methods
+## Blessed DID Methods
 
 Currently, atproto supports two DID methods:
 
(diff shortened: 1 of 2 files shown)


content/specs/did.md
```diff
@@ -76,8 +76,62 @@ did:method:val?two
...
+- Add the character `z` as a prefix, to indicate the multibase, and include no other multicodec indicators
+
+The decoding process is the same in reverse, using the curve type as context.
+
+As an internal detail of the DID PLC method, the W3C-standardized `did:key` encoding is used to represent public keys in DID PLC operations. Software only needs to process keys in this format if they are trying to validate PLC operation chains, but the details are described here anyways.
+
+This encoding includes metadata about the type of key, so they can be parsed and used unambiguously. The encoding process is:
+
+- Encode the public key curve "point" as bytes. Be sure to use the smaller "compact" or "compressed" representation. This is usually easy for `k256`, but might require a special argument or configuration for `p256` keys
+- Prepend the appropriate curve multicode indicator bytes in front of the key bytes:
```
dholms, line 125: this has been a recurring point of confusion in did:key implentations. These bytes do not match the multicodec table for the keys because they are the varint encoded version of those bytes. We just hardcoded the varint encoding

So maybe a quick note that these are the _already varint encoded_ multicodec bytes

content/specs/handle.md
```diff
@@ -92,11 +92,68 @@ laptop.local
...
+Clients can rely on network services (eg, their PDS) to resolve handles for them, using the `com.atproto.identity.resolveHandle` endpoint, and don't usually need to implement resolution directly themselves.
+
+The DNS TXT method is the recommended and preferred resolution method for individual handle configuration, but services should fully support both methods. The intended use-case for the HTTPS method is existing large-scale web services which may not have the infrastructure to automate the registration of thousands or millions of DNS TXT records.
+
+Handles should not be trusted or considered valid until the DID is also resolved and the current DID document is confirmed to link back to the handle. The link between handle and DID must be confirmed bidirectionally, otherwise anybody could create handle aliases for third-party accounts.
+
+### DNS TXT Method
+
+For this resolution method, a DNS TXT record is registered for the `_atproto` sub-domain under the handle hostname. The record value should have the prefix `did=`, followed by the full domain.
```
dholms, line 107: can reference https://datatracker.ietf.org/doc/html/rfc1464 if you like here

emilyliu7321 (approved):

PR #117, opened 2023-06-15
specs/xrpc: re-write; frame as HTTP API

bnewbold:


content/specs/xrpc.md
```diff
@@ -1,216 +1,137 @@
...
 
-Any response code not explicitly enumerated should be handled as follows:
+Only HTTPS should be used over the open internet.
 
-- `1xx` treat as a `404`
-- `2xx` treat as a `200`
-- `3xx` treat as a `404` (redirects are not supported)
-- `4xx` treat as a `400`
-- `5xx` treat as a `500`
+Care should be taken with personally identifiable information in blobs, such as EXIF metadata. It is currently the *client's* responsibility to strip any sensitive EXIF metadata from blobs before uploading.
```
dholms, line 125: Maybe add a note that a PDS _may_ strip EXIF metadata?

Seems nice for a PDS to look out for their users. I wouldn't mind putting a flag on `uploadBlob` that was something like `dontTouch` ie "don't modify these bytes at all"

emilyliu7321 (approved):

PR #118, opened 2023-06-15
specs/lexicon: re-write

bnewbold:
Put @devinivy on this one, but also cc: @dholms

**Questions for Reviewers (not part of spec doc)**

Can we say that a Lexicon *must* have at least one definition (`defs`)?

Can a `union` have an empty `refs` list, or must there always be at least one ref? What if `closed=true`? Is an empty list the same as `undefined`?

Can any types be indicated by a `ref`? What about in a `union`? How is a union disambiguate non-object types? Eg, multiple strings with different format. Nullable union and a `null` option.

Should subscription schemas *always* be a union of refs (with names)?

Why do we have `com.atproto.admin.defs#videoDetails`?

`const` and `default` are only defined for a handful of primitive types. Is this intentional?

### DIFF
--- a/content/specs/lexicon.md
+++ b/content/specs/lexicon.md
@@ -93,10 +93,10 @@ Type-specific fields:
 - `parameters` (object, optional): same as Query and Procedure
 - `message` (object, optional): specifies what messages can be
     - `description` (string, optional): short description
-    - `schema` (object, required): schema definition, either an `object`, a `ref`, or a `union` of refs
+    - `schema` (object, required): schema definition, which must be a `union` of refs
 - `errors` (array of objects, optional): same as Query and Procedure
 
-Subscription schemas are almost always a `union` of refs, not an `object` type.
+Subscription schemas (referenced by the `schema` field under `message`) must be a `union` of refs, not an `object` type.
 
 ## Field Type Definitions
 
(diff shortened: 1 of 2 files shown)


dholms:
> Can we say that a Lexicon must have at least one definition (defs)?

Yup seems fine 👍 

> Can a union have an empty refs list, or must there always be at least one ref? What if closed=true? Is an empty list the same as undefined?

I think a union can be empty & is the same as `type="unknown"`.
I'm not sure about an empty closed union 🤔 My instinct is to say that that should not be allowed

> Can any types be indicated by a ref? What about in a union? How is a union disambiguate non-object types? Eg, multiple strings with different format. Nullable union and a null option.

Right now a ref has to indicate another schema def. Therefore you can only have object-types in there - ie not primitives.
For multiple strings with different formats, I think we'd either want a convention for that or you'd need to wrap each variation in an object.

I'm not resolute in this, but that's how it works now. Can revisit if we feel it's necessary

> Should subscription schemas always be a union of refs (with names)?

Yes I believe so

> Why do we have com.atproto.admin.defs#videoDetails?

Probably just future proofing, we can likely remove for now.

> const and default are only defined for a handful of primitive types. Is this intentional?

Probably. `default` is mainly important for method params/input. Tbh I'm not sure how much we need `const`

content/specs/lexicon.md
```diff
@@ -1,313 +1,357 @@
...
+
+It should be possible to translate Lexicon schemas to JSON Schema or OpenAPI and use tools and libraries from those ecosystems to work with atproto data in JSON format.
+
+Implementations which serialize and deserialize data from JSON or CBOR in to structures derived from specific Lexicons should be aware of the risk of "clobbering" unexpected fields. For example, if a Lexicon is updated to add a new (optional) field, old implementations would not be aware of that field, and might accidentally strip the data when de-serializing and then re-serializing. Depending on the context, one way to avoid this problem is to retain any "extra" fields, or to pass-through the original data object instead of re-serializing it.
+
+## Possible Future Changes
+
+The validation rules for unexpected additional fields may change. For example, a mechanism for Lexicons to indicate that the schema is "closed" and unexpected fields are not allowed, or a convention around field name prefixes (`x-`) to indicate unofficial extension.
+
+Enums are somewhat redundant with Unions, and might be removed.
```
dholms, line 357: In my view: enums are for primitives & unions are for complex schema types

bnewbold:
@dholms:

> Right now a ref has to indicate another schema def. Therefore you can only have object-types in there - ie not primitives.

hmm, not sure I follow. `app.bsky.graph.defs#listPurpose` is a schema def with type string which is referenced by another schema. `app.bsky.graph.defs#modlist` is type token and also referenced. I do think a union can only reference an object (or higher level thing like a record), but that is a constraint on how unions work, not refs themselves, right?

re: `const`, you might already know but we do use it in a couple places in existing lexicons, so not a trivial rip-out.

content/specs/lexicon.md
bnewbold: Makes sense!

PR #119, opened 2023-06-15
specs/atp: re-write specs intro

bnewbold:

PR #120, opened 2023-06-15
(PREVIEW BRANCH) all the new specs combined

bnewbold:
**Don't review or merge this branch!**

This is just to have a combined PR preview to link and share.

Actual PRs for review:

* https://github.com/bluesky-social/atproto-website/pull/113
* https://github.com/bluesky-social/atproto-website/pull/114
* https://github.com/bluesky-social/atproto-website/pull/115
* https://github.com/bluesky-social/atproto-website/pull/116
* https://github.com/bluesky-social/atproto-website/pull/117
* https://github.com/bluesky-social/atproto-website/pull/118
* https://github.com/bluesky-social/atproto-website/pull/119


warpfork:
There's quite a few links to `www.notion.so/blueskyweb/*` which are non-public -- if those docs live in notion but are public, maybe you want to get their share links instead?  Notion isn't very good about redirecting non-logged-in users to the public versions of pages if they arrive at the internal links.

bnewbold:
@warpfork missed some, thanks! These don't link to actual documents, they were relative links between spec pages in Markdown which Notion mangled. Will resolve these in the individual PRs.

bnewbold:
TODO(bnewbold): "The first entry in the array for a given node should contain the full key, and a common prefix length of 0." => must not should

clarify that prefix compression is required and deterministic.

content/specs/data-model.md
```diff
@@ -0,0 +1,146 @@
...
+summary: Consistent data encoding for records and messages.
+---
+
+# Data Model
+
+Records and messages in atproto are stored, transmitted, encoded, and authenticated in a consistent way. The core "data model" is based on [Interplanetary Linked Data (IPLD)](https://ipld.io/docs/data-model/), a specification for hash-linked data structures from the IPFS ecosystem.
+
+When data needs to be authenticated (signed), referenced (linked by content hash), or stored efficiently, it is encoded in Concise Binary Object Representation (CBOR). CBOR is an IETF standard roughly based on JSON. IPLD specifies a normalized subset of CBOR called "DAG-CBOR", which is what atproto uses.
+
+IPLD also specifies an analogous set of conventions of JSON called "DAG-JSON", but atproto uses a different set of conventions when encoding JSON data.
```
snarfed, line 12: Thanks for this! I went off and implemented https://github.com/snarfed/dag-json a while back due to misreading the `car inspect` output and thinking the PDS itself was returning DAG-JSON. Funny, and totally PEBKAC.

content/specs/data-model.md
```diff
@@ -0,0 +1,146 @@
...
+
+In the atproto data model there is a semantic difference between explicitly setting an map field to `null` and not including the field at all. Both JSON and CBOR have the same distinction.
+
+Null or missing fields are also distinct from "false-y" value like `false` (for booleans), `0` (for integers), empty lists, or empty objects.
+
+### Why No Floats?
+
+IPLD, CBOR, and JSON all natively support floating point numbers, so why does atproto go out of the way to disallow them?
+
+The IPLD specification describes some of the complexities and sharp edges when working with floats in a content-addressable world. In short, de-serializing in to machine-native format, then later re-encoding, is not always consistent. This is definitely true for special values and corner-cases, but can even be true with "normal" float values on less-common architectures.
```
snarfed, line 53: Oh man, ugly. Good call here.

content/specs/atp.md
```diff
@@ -1,294 +1,82 @@
...
+The AT Protocol itself does not specify common social media conventions like "follows" or "avatars", leaving these to application-level Lexicons. The `com.atproto.*` Lexicons provide common APIs for things like account signup and login. These could be considered part of AT Protocol itself, though they can also be extended or replaced over time as needed. Bluesky is a Twitter-style microblogging application built on top of AT Protocol, with lexicons under the `app.bsky.*` namespace.
+
+## Cryptography
+
+Two elliptic curves are currently supported throughout the protocol, and implementations are expected to fully support both:
+
+- `p256` elliptic curve: aka "NIST P-256", aka `secp256r1` (note the `r`), aka `prime256v1`
+    - This curve *is* included in the WebCrypto API. It is commonly supported by personal device hardware (Trusted Platform Modules (TPMs) and mobile Secure Enclaves), and by cloud Hardware Security Modules (HSMs)
+- `k256` elliptic curve: aka "NIST K-256", aka `secp256k1` (note the `k`)
```
snarfed, line 33: Ah, this is useful for understanding why both are required. I think this was alluded to before somewhere, but not as explicit. Thanks.

content/specs/atp.md
```diff
@@ -1,294 +1,82 @@
...
+**Network:** HTTP client-server and server-server [APIs](/specs/xrpc) are described with Lexicons, as are WebSocket [Event Streams](/specs/event-stream). Individual records can be referenced across the network by [AT URI](/specs/at-uri-scheme). A Personal Data Server (PDS) acts as an account's trusted agent in the network, routes client network requests, and hosts repositories. A Big Graph Server (BGS) crawls many repositories and outputs a unified event firehose.
 
-## Server-to-server API
+**Application:** APIs and record schemas for applications built on atproto are specified in [Lexicons](/specs/lexicon), which are are referenced by [Namespaced Identifiers](/specs/nsid) (NSIDs). Application-specific aggregations (such as search) are provided by an Application View (AppView) service. Clients can include mobile apps, desktop software, or web interfaces.
```
snarfed, line 23: Hmm! AppView is new to me. Sounds like it overlaps a lot with BGS. Is BGS the architectural concept, and AppView is more of a service/interface?

content/specs/lexicon.md
```diff
@@ -1,313 +1,357 @@
...
+
+- `parameters` (object, optional): a schema definition with type `params`, describing the HTTP query parameters for this endpoint
+- `output` (object, optional): describes the HTTP response body
+    - `description` (string, optional): short description
+    - `encoding` (string, required): MIME type for body contents
+    - `schema` (object, required): schema definition, either an `object`, a `ref`, or a `union` of refs
+- `input` (object, optional, only for `procedure`): describes HTTP request body schema, with the same format as the `output` field
+- `errors` (array of objects, optional): set of string error codes which might be returned
+    - `name` (string, required): short name for the error type, with no whitespace
+    - `descriptoin` (string, optional): short description, one or two sentences
```
snarfed, line 87: minor nit, typo, descriptoin

content/specs/lexicon.md
```diff
@@ -1,313 +1,357 @@
...
+
+## Usage and Implementation Guidelines
+
+It should be possible to translate Lexicon schemas to JSON Schema or OpenAPI and use tools and libraries from those ecosystems to work with atproto data in JSON format.
+
+Implementations which serialize and deserialize data from JSON or CBOR in to structures derived from specific Lexicons should be aware of the risk of "clobbering" unexpected fields. For example, if a Lexicon is updated to add a new (optional) field, old implementations would not be aware of that field, and might accidentally strip the data when de-serializing and then re-serializing. Depending on the context, one way to avoid this problem is to retain any "extra" fields, or to pass-through the original data object instead of re-serializing it.
+
+## Possible Future Changes
+
+The validation rules for unexpected additional fields may change. For example, a mechanism for Lexicons to indicate that the schema is "closed" and unexpected fields are not allowed, or a convention around field name prefixes (`x-`) to indicate unofficial extension.
```
snarfed, line 355: Just FYI, the standards community has generally moved away from experimental X- prefixes, and agreed that htey caused more harm than good. https://datatracker.ietf.org/doc/html/rfc6648#appendix-B . Same with vendor prefixes more recently, https://www.webstandards.org/2012/02/09/call-for-action-on-vendor-prefixes/index.html

content/specs/lexicon.md
```diff
@@ -1,313 +1,357 @@
...
+
+Protocol implementations should generally consider data which fails to validate against the Lexicon to be entirely invalid, and should not try to repair or do partial processing on the individual piece of data.
+
+Unexpected fields in data which otherwise conforms to the Lexicon should be ignored. When doing schema validation, they should be treated at worst as warnings. This is necessary to allow evolution of the schema by the controlling authority, and to be robust in the case of out-of-date Lexicons.
+
+Third parties can technically insert any additional fields they want in to data. This is not the recommended way to extend applications, but it is not specifically disallowed. One danger with this is that the Lexicon may be updated to include fields with the same field names but different types, which would make existing data invalid.
```
snarfed, line 345: 💯

snarfed: ...also one thing that's missing here is guidance on whether PDSes should accept records with unknown lexicons, and if so, how they should validate them, etc. Depends on lexicon resolution I guess?

content/specs/handle.md
```diff
@@ -92,11 +92,68 @@ laptop.local
...
+
+For this resolution method, a DNS TXT record is registered for the `_atproto` sub-domain under the handle hostname. The record value should have the prefix `did=`, followed by the full domain.
+
+For example, the handle `bsky.app` would have a TXT record on the name `_atproto.bsky.app`, and the value would look like `did=did:plc:z72i7hdynmk6r22z27h6tvur`.
+
+Any TXT records with values not starting with `did=` should be ignored. Only a single valid record should exist at any point in time. If multiple valid records with different DIDs are present, resolution should fail. In this case resolution can be re-tried after a delay, or using a recursive resolver.
+
+Note that very long handles can not be resolved using this method if the additional `_atproto.` name segment pushes the overall name over the 253 character maximum for DNS queries. The HTTPS method will work for such handles.
+
+DNSSEC is not required.
```
snarfed, line 115: 👍 😭

content/specs/lexicon.md
snarfed: @pfrazee implied/said yes in https://github.com/bluesky-social/atproto/discussions/855 :
> the goal is to make the PDS indifferent to application schemas and provide generic behaviors only; that way adding some new schema doesnt run the risk of losing forward progress

content/specs/xrpc.md
```diff
@@ -1,216 +1,137 @@
...
+## Authentication
 
-Method schemas are designed to be machine-readable and network-accessible. While it is not currently _required_ that a schema is available on the network, it is strongly advised to publish schemas so that a single canonical & authoritative representation is available to consumers of the method.
+The current authentication scheme for XRPC is HTTP Bearer auth with JWT tokens, including refresh tokens. This scheme is likely to be extended or replaced.
 
-To fetch a schema, a request must be sent to the builtin [`getSchema`](#getschema) method. This request is sent to the authority of the NSID.
+Not all endpoints require authentication, but there is not yet a consistent way to enumerate which endpoints do or do not.
 
-### Built-in methods
+Initial login uses the `com.atproto.server.createSession` endpoint, including the password and an accoun tidentifier (eg, handle or registered email address). This returns a `refreshJwt` (as well as an initial `accessJwt`).
```
snarfed, line 51: minor nit, typo, accoun tidentifier

content/specs/handle.md
```diff
@@ -92,11 +92,68 @@ laptop.local
 blah.arpa
 ```
 
-### Usage and Implementation Guidelines
+## Handle Resolution
```
snarfed, line 95: Thanks! This is all great.

As mentioned before, definitely questions here around how this interacts with federation, but that's for the future. And ideally guidance on when/how to deal with handles in records as alternatives to DIDs.

content/specs/xrpc.md
```diff
@@ -1,216 +1,137 @@
...
+App Passwords are a mechanism to reduce security risks when logging in to third-party clients and web applications. Accounts can create and revoke app passwords separate from their primary password. They are used to log in the same way as the primary password, but grant slightly restricted permissions to the client application, preventing destructive actions like account or changes to authentication settings (including app passwords themselves).
 
-The HTTP Method used depends on the `type` specified by the method schema.
+Clients and apps themselves do not need to do anything special to use app passwords. It is a best practice for most clients and apps to include a reminder to use an app password when logging in. App passwords usually have the form `xxxx-xxxx-xxxx-xxxx`, and clients can check against this format to prevent accidental logins with primary passwords (unless the primary password itself has this format).
 
-|Type|Method|
-|-|-|
-|`query`|`GET`|
-|`procedure`|`POST`|
+### Admin Token
```
snarfed, line 63: Interesting! I'm guessing this is primarily a legacy holdover, so it's more descriptive of the current prod PDS than normative, right? Otherwise it seems like this could be an implementation detail that doesn't really need to standardized, or if it does, maybe it would fit better in the protocol or repository docs?

content/specs/event-stream.md
```diff
@@ -0,0 +1,133 @@
...
+- `error` (string, required): the error type name, with no namespace or `#` prefix
+- `message` (string, optional): a description of the error
+
+Streams should be closed immediately following transmitting or receiving an error frame.
+
+Message payloads must always be objects. They should omit the `$type` field, as this information is already indicated in the header. There is no specific limit on the size of WebSocket frames in atproto, but they should be kept reasonably small. A couple megabytes is an informal guideline.
+
+If a client can not keep up with the rate of messages, the server may send a "too slow" error and close the connection.
+
+### Sequence Numbers
```
snarfed, line 68: Oof, this is a lot. Sure would be nice if we could get some of it for free by switching the namespace from PDS hostname + endpoint to user DID + collection and then depending on monotonically increasing TIDs instead. I guess that would mean servers and clients would have to keep track of O(users) sequence TIDs instead of just O(servers/clients + endpoints) sequence numbers though.

content/specs/atp.md
devinivy: The BGS is primarily a "dumb" rehoster of content across PDSes, whereas the AppView provides the views specific to the Bluesky application.  The BGS doesn't need to have semantic awareness of the content it's hosting, whereas the AppView is interpreting "like", "follow", "post" records into more complex views to service an application.  Roughly speaking, BGS is concerned with `com.atproto.sync.*` and `com.atproto.repo.*` lexicons, and the Bluesky AppView is concerned with `app.bsky.*` lexicons.

content/specs/lexicon.md
bnewbold: Thanks for the catch!

content/specs/lexicon.md
bnewbold: Thanks for the refs! We do need to think through what the best practice is here better. I guess we think of the primary extension mechanism (new lexicons) as being pretty clear, but that might be because we are in a weird perspective of having full control over the most popular lexicons. other folks sure do seem to want to stuff fields in these lexicons, and we should have a better answer.

content/specs/lexicon.md
bnewbold: Yeah, we definitely want PDS implementations to be more agnostic to the lexicons of data they host. Right now, in prod, we still have PDS+AppView conjoined, which is why we haven't loosened up on this yet.

We wanted to have a bit more confidence in this by first actually deploying federation with the `app.bsky` lexicons working with PDS+BGS+AppView as distinct network services (happening as a slow-roll transition right now), and probably us ourselves doing an example separate/independent lexicon as a secondary app to see how that goes, before we are confident in this entire architecture. That does take time and we should probably just get out of the way and let devs experiment in other spaces though. One step would be to allow random un-validated records outside the `com.atproto` and `app.bsky` spaces, to keep those on a shorter leash to start.

We also haven't worked out the "fetch Lexicon automatically over the network given the NSID". A couple options on how to do that, none super complex, but not clear if that will be reliable enough to require fetching and validation for new record types, or to end up with a schema where PDS instances have some "popular" or "common" set of lexicons built-in (compiled-in?) and work with just those.

content/specs/handle.md
bnewbold: Yeah... sort of poking the hornets nest there.

maybe we should require IPv6 though? :smiling_imp:

content/specs/xrpc.md
bnewbold: thanks!

content/specs/xrpc.md
bnewbold: Yeah, all the auth stuff is sort of temporary and we expect to improve a lot on this. But that isn't going to be a quick fix, and we want to be transparent and clear how the existing scheme works in the meanwhile.

We do expect to have some of the other things, like repo history and sync stream, firmed up sooner than auth, which is why we just don't include those at all yet.

content/specs/event-stream.md
bnewbold: This is under pretty active discussion internally, and with a few external tech advisors. Hard to compress all the conversation to a short summary. Also, realistically, this is fairly high up in the protocol stack and the sort of thing that might have multiple implementations over time. Eg, internally and at scale, can't really imagine not using Kafka for the big firehoses; maybe we'll have a formal "how to do this in Kafka within an org/consortium" spec or recommendation.

bnewbold: Also, clarified that the sequence stuff is optional. Can totally create subscription Lexicons which don't include it.

bnewbold:
Thanks for feedback from everybody on this group of changes!

Everything in thing branch has now been merged and deployed in to the live website. Future corrections should go in new issues or PRs.

There are some active discussions here, and this branch was linked from a couple places. I'll leave it for another day or two then close the PR (which IIUC will remove the PR preview site).

PR #121, opened 2023-06-16
Add blue_factory Ruby gem to community projects

mackuba:
It's a feed generator for custom feeds implemented in Ruby using Sinatra.


emilyliu7321 (approved):
thanks!

PR #123, opened 2023-06-17
cryptography: note about ECDSA sigs and low-S requirement

bnewbold:
This has come up a couple times, think I finally found the right jargon to describe the issue.

I'm not actually super confident in the requirement here, but the typescript reference implementation *does* reject low-S signatures (apparently), so seems best to require them broadly for consistency? If we decide to go ahead with this, somebody (can be me) should add tests for all of our implementations (typescript, golang, did-plc) to ensure we are consistent.

cc: @whyrusleeping in case you have an opinion


dholms (approved):
yup yup this is correct & intentional

There's not a big attack surface here, I suppose the attack would be gossiping around a different commit hash which could look like a fork of the repo & maybe cause a bit of churn in indexing behavior

I can't think of any reason not to lock this down & it just makes everything a bit tighter.

Altho this does remind me: I should check what web crypto does for p256 sigs 🤔

PR #126, opened 2023-06-18
specs/xrpc: inter-service JWT auth

bnewbold:
Based on comments from Slack. I haven't implemented this myself, so not super confident, but this is also just a temporary scheme.

Also indicated that the admin token scheme is probably temporary.

### DIFF
--- a/content/specs/xrpc.md
+++ b/content/specs/xrpc.md
@@ -60,7 +60,7 @@ App Passwords are a mechanism to reduce security risks when logging in to third-
 
 Clients and apps themselves do not need to do anything special to use app passwords. It is a best practice for most clients and apps to include a reminder to use an app password when logging in. App passwords usually have the form `xxxx-xxxx-xxxx-xxxx`, and clients can check against this format to prevent accidental logins with primary passwords (unless the primary password itself has this format).
 
-### Admin Token
+### Admin Token (Temporary Specification)
 
 Some administrative XRPC endpoints require authentication with admin privileges. The current scheme for this is to use HTTP Basic authentication with user "admin" and a fixed token in the password field, instead of HTTP Bearer auth with a JWT. This means that admin requests do not have a link to the account or identity of the client beyond "admin".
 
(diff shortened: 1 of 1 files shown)


content/specs/xrpc.md
```diff
@@ -78,6 +78,30 @@ The set of endpoints requiring admin auth is likely to get out of date in this s
...
+- `iss` body field: account DID that the request is being sent on behalf of
+- `aud` body field: service DID associated with the service that the request is being sent to
+- `exp` body field: token expiration time, as a UNIX timestamp with seconds precision. Should be a short time window, as revocation is not implemented. 60 seconds is a good value.
+- JWT signature: base64url-encoded signature using the account DID's signing key
+
+The signature is computed using the regular JWT process. As Typescript pseudo-code, this looks like:
+
+```
+const headerPayload = utf8ToBase64Url(jsonStringify(header)) + '.' + utf8ToBase64Url(jsonString(body))
+const signature = sign(sha256(utf8Bytes(headerPayload)))
```
dholms, line 101: my bad on not including it in the pseudo code - but maybe clarify in this function that `sign` is done with the user's atproto signing key - same as a repo signature

dholms (approved):
just the one clarification - which we have in text above, but might be good in the pseudo code as well

besides that looks great!

content/specs/xrpc.md
devinivy: Sort of related, I personally think of the `sha256()` as being part of the signing process.  At the same time I appreciate the explicitness here about how to generate the signature.  I think the risk would be someone double-hashing by mistake, not realizing that their keypair might be doing that on its own, so if there's a way to make a clarification to avoid that it could be useful.  But also not a huge deal and probably clear enough for most implementers, not terribly worried about this.

bnewbold: My feeling on this is that in our SDKs we should call the primary function `hash_and_sign()`. I think it is equally confusing to have a convention that `sign()` also implicitly hashes.

A python-ic alternative would be something like `sign(raw_data, hash="sha256")` as an (optional) keyword arg.

dholms: yeah I agree with these & we probably shouldn't explicilty say `sha256` here

I'd be game for saying `hashAndSign` in the pseudo-code & we can look at adjusting it in the SDK as well

I've found there is annoyingly not a strong convention in crypto libs around if the hash is part of the signing process :/

bnewbold: The pair of `hashAndSign` is also `hashAndVerify`, which sounds even worse, but has also bitten me a bunch as well. Have thought about these a bunch in the context of having relatively consistent SDK API shape across programming languages.

PR #140, opened 2023-07-07
Fix links to specs in Identity guide

caskuplich:
Fix links to Handles and DIDs specs in Identity guide. The text shows markdown links syntax instead the rendered links. This pull request fixes them with HTML `<a>` tags.


bnewbold (approved):
Thanks!

PR #143, opened 2023-07-08
Update projects to include ScoreMyWordle bot

shaneafsar:
Added an entry for ScoreMyWordle, which is a bot that now runs across Bluesky, Mastodon, and Twitter.


emilyliu7321 (approved):
thanks!

PR #149, opened 2023-07-25
Added Bossett's feedgen to Custom Feeds

Bossett:
Added my feedgen, which is basically an expanded version of the official one.

Not sure if there is formatting etiquette (i.e. two links, link ordering, etc.)?


emilyliu7321 (approved):
thank you!

Issue #128, opened 2023-06-19
specs/lexicon: document new 'language' string format


bnewbold:
As added in https://github.com/bluesky-social/atproto/pull/1219

PR #152, opened 2023-07-26
specs/lexicon: document 'language' string format

bnewbold:
Finally getting around to this. Tagging Devin on this one as he did the typescript implementation (back in late June).

closes: https://github.com/bluesky-social/atproto-website/issues/128

### DIFF
--- a/content/specs/lexicon.md
+++ b/content/specs/lexicon.md
@@ -318,9 +318,9 @@ Maximum length in Lexicons is 8 KBytes.
 
 ### `language`
 
-An [IETF Language Tag](https://en.wikipedia.org/wiki/IETF_language_tag) string, compliant with [BCP 47](https://www.rfc-editor.org/info/bcp47), defined in [RFC 5646](https://www.rfc-editor.org/rfc/rfc5646.txt) ("Tags for Identifying Languages"). This is the same standard used to identify languages in HTTP, HTML, and other web standards.
+An [IETF Language Tag](https://en.wikipedia.org/wiki/IETF_language_tag) string, compliant with [BCP 47](https://www.rfc-editor.org/info/bcp47), defined in [RFC 5646](https://www.rfc-editor.org/rfc/rfc5646.txt) ("Tags for Identifying Languages"). This is the same standard used to identify languages in HTTP, HTML, and other web standards. The Lexicon string must validate as a "well-formated" language tag, as defined in the RFC. Clients should ignore language strings which are "well-formed" but not "valid" according to the RFC.
 
-ISO 639 two-character and three-character language codes can be used on their own, lower-cased, such as `ja` (Japanese) or `ban` (Balinese). Regional sub-tags can be added, like `pt-BR` (Brazilian Portuguese). Additional subtags can also be added, such as `hy-Latn-IT-arevela`.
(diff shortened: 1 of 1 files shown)


content/specs/lexicon.md
```diff
@@ -315,6 +316,14 @@ Invalid examples:
 Flexible to any URI schema, following the generic RFC-3986 on URIs. This includes, but isn’t limited to: `did`, `https`, `wss`, `ipfs` (for CIDs), `dns`, and of course `at`.
 Maximum length in Lexicons is 8 KBytes.
 
+### `language`
+
+An [IETF Language Tag](https://en.wikipedia.org/wiki/IETF_language_tag) string, compliant with [BCP 47](https://www.rfc-editor.org/info/bcp47), defined in [RFC 5646](https://www.rfc-editor.org/rfc/rfc5646.txt) ("Tags for Identifying Languages"). This is the same standard used to identify languages in HTTP, HTML, and other web standards.
```
devinivy, line 321: One little note here— the spec has two classes of conformance ([link](https://www.rfc-editor.org/rfc/rfc5646.html#section-2.2.9)): "well-formed" and "valid".  One of the requirements for valid tags is that they align with the living IANA language subtag registry.  For that reason, our current validation only looks for well-formedness.  I don't think that's set in stone, but that's the current state of things 👍 

```
   A tag is considered "well-formed" if it conforms to the ABNF
   Section 2.1.  Language tags may be well-formed in terms of syntax
   but not valid in terms of content.  However, many operations
   involving language tags work well without knowing anything about the
   meaning or validity of the subtags.
```

bnewbold: I think we should stick with "well-formed" as the string format requirement, a say that folks should ignore languages which are "well-formed" but not "valid". My read is that "valid" will expand over time (with additional language codes, suffixes, etc), but "well-formed" is more stable.

devinivy: That sounds good 👍

content/specs/lexicon.md
```diff
@@ -315,6 +316,14 @@ Invalid examples:
 Flexible to any URI schema, following the generic RFC-3986 on URIs. This includes, but isn’t limited to: `did`, `https`, `wss`, `ipfs` (for CIDs), `dns`, and of course `at`.
 Maximum length in Lexicons is 8 KBytes.
 
+### `language`
+
+An [IETF Language Tag](https://en.wikipedia.org/wiki/IETF_language_tag) string, compliant with [BCP 47](https://www.rfc-editor.org/info/bcp47), defined in [RFC 5646](https://www.rfc-editor.org/rfc/rfc5646.txt) ("Tags for Identifying Languages"). This is the same standard used to identify languages in HTTP, HTML, and other web standards. The Lexicon string must validate as a "well-formated" language tag, as defined in the RFC. Clients should ignore language strings which are "well-formed" but not "valid" according to the RFC.
```
devinivy, line 321:
```suggestion
An [IETF Language Tag](https://en.wikipedia.org/wiki/IETF_language_tag) string, compliant with [BCP 47](https://www.rfc-editor.org/info/bcp47), defined in [RFC 5646](https://www.rfc-editor.org/rfc/rfc5646.txt) ("Tags for Identifying Languages"). This is the same standard used to identify languages in HTTP, HTML, and other web standards. The Lexicon string must validate as a "well-formed" language tag, as defined in the RFC. Clients should ignore language strings which are "well-formed" but not "valid" according to the RFC.
```

bnewbold:
@devinivy is this looking good to go now?

dholms (approved):
this looks good to me 👍 

but @devinivy is the one that got most into this so I'll wait for his signoff 👌

devinivy (approved):
Looks great!

PR #154, opened 2023-07-27
specs/handle: document handle.invalid special value

bnewbold:
Short addition to have something in here when this feature goes live (soon now).

Could expand on the semantics a bit more.

cc: @pfrazee who discussed this recently in https://github.com/bluesky-social/atproto/issues/1362

### DIFF
--- a/content/specs/handle.md
+++ b/content/specs/handle.md
@@ -47,12 +47,13 @@ To summarize the above, the initial list of disallowed TLDs includes:
 
 - `.local`
 - `.arpa`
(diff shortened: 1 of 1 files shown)


bnewbold:
@dholms gentle nudge on this PR

dholms (approved):
yup 👍 

If interested, could link to prior art in the [Session Initiation Protocol](https://en.wikipedia.org/wiki/.invalid). But not sure that's necessary

PR #155, opened 2023-07-29
lexicon spec doc bug fix: schema is ignored for non-JSON encodings

snarfed:
Minor fix. I think this is right, but feel free to correct me if not. Thanks in advance!

### DIFF
--- a/content/specs/lexicon.md
+++ b/content/specs/lexicon.md
@@ -83,7 +83,7 @@ Type-specific fields:
 - `output` (object, optional): describes the HTTP response body
     - `description` (string, optional): short description
     - `encoding` (string, required): MIME type for body contents
-    - `schema` (object, required): schema definition, either an `object`, a `ref`, or a `union` of refs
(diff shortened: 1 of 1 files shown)


bnewbold:
Thanks for the catch!

I think this might actually just be `optional` all the time. At least, that is how it is in the typescript implementation. I think it should be possible to have arbitrary (unspec'd) `application/json` response endpoints.

snarfed:
Sure! You know best, happy to update this PR or close, whichever's easiest.

bnewbold:
I pulled in your commit to retain history (thanks for the catch/contribution!), but am closing this in favor of https://github.com/bluesky-social/atproto-website/pull/177 (some tweaking of language). Appologies if there was a more elegant git/github way to wrangle this.

snarfed:
Hey, sounds great! I'm not at all worried about git credit, but thank you, appreciate it. Glad to see the update.

PR #160, opened 2023-08-03
add admonitions, postcss nesting

estrattonbailey:
Adds "admonitions", popularized by Docusaurus, using Remark's own `remark-directive`. This PR adds four types we can use: `note`, `tip`, `warn` and `danger`.

Example:

```markdown
# Document Title

Some text...

:::tip
**Bold works well for title**
But within this, you can write any markdown you want.
:::
```

We may need to adjust styles depending on needs, but it's a pretty simple approach overall.

Screen Shot 2023-08-03 at 2 57 03 PM


bnewbold (approved):
These look great to me!

ericvolp12 (approved):
LGTM, awesome, thanks!

PR #164, opened 2023-08-04
specs/did: note about P-256 verificationMethod

bnewbold:
Short, hopefully temporary, note about P-256 `verificationMethod` type not being totally above-board.


dholms (approved):
looks good 👍

PR #175, opened 2023-08-14
Fix NSID regex.

aDotInTheVoid:
Copied from https://github.com/bluesky-social/atproto/blob/73f64009a22b3428a21a6d7807aee89d629fd0a4/packages/nsid/src/index.ts#L100

This now matches `cn.8.lex.stuff`, which the old regex didn't.

Not sure why the code used `[a-zA-Z]([a-zA-Z]{0,61}[a-zA-Z])?` instead of `[a-zA-Z]{1,63}`, but I don't think it's worth changing in the `atproto` repo given limited review bandwidth, and I'd rather this regex was exactly the same as their, than have it be slightly more concise here.

### DIFF
--- a/content/specs/nsid.md
+++ b/content/specs/nsid.md
@@ -41,7 +41,7 @@ The comprehensive list of syntax rules is:
 A reference regex for NSID is:
 
 ```
-/^[a-zA-Z]([a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(\.[a-zA-Z]([a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)+(\.([a-zA-Z]{1,63}))$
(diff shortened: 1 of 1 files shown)


bnewbold:
Thank you so much for catching this!

The spec was updated more recently than the atproto implementation. I can only imagine that I tried to simplify things in the spec doc (the `{1,63}` change; the more verbose syntax is probably from copy-pasting and editing the earlier component regex patterns), and then made a typo.

Simplifying the regex used in code, and here, would be nice eventually, but that specific code has some unrelated open PRs and I don't want to create conflicts, so agree with updating here for consistency.

I double checked and this regex now exactly matches that in the current atproto `main` branch (`ea9d96e3a44119ca6189e7a3989a1bd9b54989a9`).

PR #177, opened 2023-08-15
specs/lexicon: clarify `schema` being optional on outputs

bnewbold:
This is an additional patch on top of https://github.com/bluesky-social/atproto-website/pull/155

As noted in the earlier PR:

> I think this might actually just be optional all the time. At least, that is how it is in the typescript implementation. I think it should be possible to have arbitrary (unspec'd) application/json response endpoints.

cc: @devinivy for a quick review (two-line specs commit)


devinivy (approved):
Yep yep! 👍

PR #178, opened 2023-08-16
specs: update DID document and crypto key encoding

bnewbold:
DO NOT MERGE YET!

This is prep for changes to DID Document formatting, key encoding, and use of `Multikey` type in `verifiedMethod` objects. Also moves key encoding info in to the crypto page (out of DID), except for the special-case legacy format (which will get deprecated eventually).

cc: @dholms

### DIFF
--- a/content/specs/cryptography.md
+++ b/content/specs/cryptography.md
@@ -40,7 +40,7 @@ When encoding public keys as strings, the preferred representation uses multibas
 - Prepend the appropriate curve multicodec value, as varint-encoded bytes, in front of the key bytes:
     - `p256` (compressed, 33 byte key length): `p256-pub`, code 0x1200, varint-encoded bytes: [0x80, 0x24]
     - `k256` (compressed, 33 byte key length): `secp256k1-pub`, code 0xE7, varint bytes: [0xE7, 0x01]
-- Encode the combined bytes with `base58btc`, yielding a string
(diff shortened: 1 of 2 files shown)


content/specs/cryptography.md
```diff
@@ -32,6 +32,28 @@ In atproto, use of the "low-S" signature variant is required for both `p256` and
...
-### Possible Future Changes
+## Public Key Encoding
+
+When encoding public keys as strings, the preferred representation uses multibase (with `base58btc` specifically) and a multicode prefix to indicate the specific key type. By embedding metadata about the type of key in the encoding itself, they can be parsed unambiguously. The process for encoding a public key in this format is:
+
+- Encode the public key curve "point" as bytes. Be sure to use the smaller "compact" or "compressed" representation. This is usually easy for `k256`, but might require a special argument or configuration for `p256` keys
+- Prepend the appropriate curve multicodec value, as varint-encoded bytes, in front of the key bytes:
+    - `p256` (compressed, 33 byte key length): `p256-pub`, code 0x1200, varint-encoded bytes: [0x80, 0x24]
+    - `k256` (compressed, 33 byte key length): `secp256k1-pub`, code 0xE7, varint bytes: [0xE7, 0x01]
+- Encode the combined bytes with `base58btc`, yielding a string
```
dholms, line 43: and prefix with `z`

bnewbold: good catch. also added example encoded keys just below

PR #211, opened 2023-10-03
specs: clarifications with datetime format

bnewbold:
Some clarifications and extra language around datetime.

Left some ambiguity about "earliest supported date" until we have consensus on that, but i'm tempted to stay that datetimes before UNIX epoch are just not allowed, as out-of-scope for what we intend this field to be used for. Eg, we might want to convert `createdAt` to a UNIX timestamps as an internal thing in various databases, and that won't work with older dates. Maybe that is too unexpected/conservative though.


devinivy (approved):
💎

PR #224, opened 2023-10-13
fix app password link

JustinChavez:
I adjusted the link because the original fragment identifier no longer points to a valid location. It now directs readers to the current App Password documentation.

This has been tested locally.


bnewbold:
Thanks!

PR #236, opened 2023-10-28
Added Greenland to Android clients

itsWindows11:


emilyliu7321 (approved):
nice name 😅 , excited to see where you take this!

itsWindows11:
I wanted a unique name for a Bluesky client since everyone else is naming it similarly to skies, but the name would be something that's still somewhat related to a blue sky. That's where I came up with this 😅

PR #247, opened 2023-11-09
Ambiguity in XRPC pagination section

ShreyanJain9:
I think this may have been a typo?


pfrazee (approved):
Yeah that looks correct to me

bnewbold (approved):
indeed, typo. thanks for the sharp eyes, and following up!

Issue #171, opened 2023-08-12
specs/lexicon: add recordkey (`rkey`) and TID (`tid`) string format types


bnewbold:
This is mostly just symmetry/completeness. We specify the syntax of these in the specs, but don't have Lexicon-specific formats for them. `rkey` is pretty flexible, but has a restricted character set and doesn't allow whitespace.

`tid` isn't used in any current Lexicons, but has a more restricted syntax, and may pop up in more sync Lexicons if we use it for repo rev/clock field.

I can both draft a spec PR and add syntax check tests to atproto (typescript) if that is helpful.

cc: @devinivy @dholms

PR #248, opened 2023-11-10
specs: add tid and record-key string formats

bnewbold:
Will probably wait until support lands in atproto (typescript) before merging this.

closes: https://github.com/bluesky-social/atproto-website/issues/171


bnewbold:
I think we have consensus/signoff on this as a feature, just waiting until the typescript reference implementation is updated before merging to specs. Will also want to do a pass over existing Lexicons adding these labels.

PR #284, opened 2024-02-04
Fix hero button text so it doesn't wrap

Jacob2161:
We can't live like this right? :grin: 

image

->

image

(been bugging me for a while)


bnewbold:
I can't repro this in my browser (chromium) at any desktop width. It should fold to vertically -stacked buttons at narrow screen width.

Not against merging if it solves problems for other browsers though.

bnewbold:

PR #289, opened 2024-02-08
Update blueskyweb.xyz links to equivalent pages on bsky.social

surfdude29:
This PR updates three links that currently point to `blueskyweb.xyz` to instead point to the equivalent pages on `bsky.social`.


bnewbold (approved):
thanks!

PR #291, opened 2024-02-13
specs: v3 commit prev as nullable but non-optional

bnewbold:
Resolve compatibility issues with "nullable and optional" fields in this core data structure with golang.

Background here: https://github.com/bluesky-social/indigo/issues/327

Basically signatures fail to validate in golang if the field is not included but `null`.

It would have been possible to just remove this field from the spec, or have it be optional (but not nullable) and those would have worked fine with golang, but the boat has basically sailed (tons of existing data in this format). So we'll stick with requiring the field exist, but be nullable.


dholms (approved):
🙌

PR #311, opened 2024-05-10
specs: remove caveat around proxying

bnewbold:
Recently merged the change that XRPC proxying is more generic, so can remove the disclaimer from the specs.


bnewbold:
This shipped a while back, and these are just removals of small disclaimers. I'm going to self-merge.

PR #316, opened 2024-06-24
do not need www for docs site

bnewbold:


bnewbold:
these links are prominent and broken, i'm going to self-merge

PR #324, opened 2024-08-07
Fix a minor typo in the documentation

Gorcenski:
In [the `#params` section](https://atproto.com/specs/lexicon#params) of the Lexicon documentation there was a typo where `procedure` was spelled `procedue`. Since this typo is formatted as code, it's probably best to ensure it's correct.


gaearon (approved):
thx!

PR #335, opened 2024-09-10
Fix incorrect link in image card

bam1279:
Hello,

I have corrected the link in the image card. It was mistakenly pointing to "https://arxiv.org/abs/2402.03239" instead of the intended "https://steveklabnik.com/writing/how-does-bluesky-work/".

Thank you for reviewing this change.

I would also like to thank you for the recent addition of the Japanese translation on atproto.com!


pfrazee (approved):
Very kind of you! Thank you!

PR #337, opened 2024-09-13
Update status of community Python SDK

MarshalX:
I hope I understand the meaning of these statuses right. The Python SDK status is pretty easy: many client things are here, but no server things (or so minimum as no) @pfrazee


pfrazee (approved):
Great! I'll take your word for it <3

PR #343, opened 2024-09-26
move specs/auth to specs/oauth

bnewbold:
This single specs page is already huge, shouldn't try to bundle other auth stuff in there as well. Best to rip this bandaid sooner than later.

This PR includes a redirect and internal updates; I can do quick updates of links in other places (docs site, blog posts) after this is deployed.


matthieusieben (approved):
Good point 👍

PR #345, opened 2024-09-26
specs/data: reserve $-prefixed field names

bnewbold:
Closes: https://github.com/bluesky-social/atproto-website/issues/170

I think this covers the basic rule and motivation. We could also mention from Lexicon spec? But this feels like the best spot.

### DIFF
--- a/src/app/[locale]/specs/data-model/page.mdx
+++ b/src/app/[locale]/specs/data-model/page.mdx
@@ -45,7 +45,7 @@ Lexicons can include additional validation constraints on individual fields. For
 
 Lexicon string fields can have additional `format` type information associated with them for validation, but as with other validation constraints this information is not available without the Lexicon itself.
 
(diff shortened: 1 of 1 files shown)


src/app/[locale]/specs/data-model/page.mdx
```diff
@@ -45,6 +45,8 @@ Lexicons can include additional validation constraints on individual fields. For
 
 Lexicon string fields can have additional `format` type information associated with them for validation, but as with other validation constraints this information is not available without the Lexicon itself.
 
+Data field names starting with `$` are reserved for use by the data model or protocol itself, in both JSON and CBOR representations. For example, the `$bytes` key name (used in CBOR and JSON), the `$link` key (used for JSON CID Links), or `$type` (used to indicate record type). Implementations should ignore unknown `$` fields (to allow protocol evolution). Applications, extensions, and integrations should not use or unilaterally define new `$` fields, to prevent conflicts as protocol evolves.
```
Comment (devinivy) [inline src/app/[locale]/specs/data-model/page.mdx:48]:
```suggestion
Data field names starting with `$` are reserved for use by the data model or protocol itself, in both JSON and CBOR representations. For example, the `$bytes` key name (used in CBOR and JSON), the `$link` key (used for JSON CID Links), or `$type` (used to indicate record type). Implementations should ignore unknown `$` fields (to allow protocol evolution). Applications, extensions, and integrations should not use or unilaterally define new `$` fields, to prevent conflicts as the protocol evolves.
```

devinivy (approved):

Issue #327, opened 2024-08-27
specs: update service auth and session auth JWT docs with updates


bnewbold:
Ref: https://github.com/bluesky-social/atproto/pull/2743

- `iat` in service auth (token created at)
- `typ` header field distinguishes between `at+jwt` ("access token") and `refresh+jwt` (refresh)

matthieusieben:
Also: `typ` of service auth would be set to `JWT`

PR #347, opened 2024-09-26
specs/xrpc: more JWT fields; document typ

bnewbold:
Pulls in the `lxm` and `jti` updates from: https://github.com/bluesky-social/atproto/discussions/2687

And the `iat` and `jti` fields, and `typ` values.

Please review the types (string/number/integer) and whether required or not. I'm not really sure how to specify the data type of, eg, `iat`: it is JSON so can't really say "integer", right? Just numeric? If these are floats, is it expected that they be validated with sub-second precision?

Trying to keep this PR tight/simple, so not reviewing/touching any "should" / "must" type language elsewhere in these sections.

Closes: https://github.com/bluesky-social/atproto-website/issues/327

cc: @rudyfraser doing PDS work and generating inter-service auth JWTs

### DIFF
--- a/src/app/[locale]/specs/xrpc/page.mdx
+++ b/src/app/[locale]/specs/xrpc/page.mdx
@@ -92,7 +92,7 @@ The JWT parameters are:
 - `alg` header field (string, required): indicates the signing key type (see [Cryptography](/specs/cryptography))
     - use `ES256K` for `k256` keys
     - use `ES256` for `p256` keys
-- `typ` header field (string, required): `JWT`
+- `typ` header field (string, required): currently `JWT`, but intend to update to a more specific value
 - `iss` body field (string, required): account DID that the request is being sent on behalf of. This may include a suffix service identifier; see below
 - `aud` body field (string, required): service DID associated with the service that the request is being sent to
 - `exp` body field (number, required): token expiration time, as a UNIX timestamp with seconds precision. Should be a short time window, as revocation is not implemented. 60 seconds is a good token lifetime.
--- a/src/app/[locale]/specs/xrpc/page.mdx
+++ b/src/app/[locale]/specs/xrpc/page.mdx
@@ -53,7 +53,9 @@ There is also a legacy authentication scheme using HTTP Bearer auth with JWT tok
 
 Most requests should be authenticated using an access JWT, but the validity lifetime for these tokens is short. Every couple minutes, a new access JWT can be requested by hitting the `com.atproto.server.refreshSession` endpoint, using the refresh JWT instead of an access JWT.
 
-The JWTs themselves should be treated as opaque tokens.
+Clients should treat the tokens as opaque string tokens: the JWT fields and semantics are not a stable part of the specification.
+
+Servers (eg, PDS implementations) which generate and valiate auth JWTs should implement domain separation between access and refresh tokens, using the `typ` header field: access tokens should use `at+jwt`, and refresh tokens should use `refresh+jwt`.
 
 ### App Passwords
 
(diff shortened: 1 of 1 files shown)


src/app/[locale]/specs/xrpc/page.mdx
```diff
@@ -53,7 +53,9 @@ There is also a legacy authentication scheme using HTTP Bearer auth with JWT tok
 
 Most requests should be authenticated using an access JWT, but the validity lifetime for these tokens is short. Every couple minutes, a new access JWT can be requested by hitting the `com.atproto.server.refreshSession` endpoint, using the refresh JWT instead of an access JWT.
 
-The JWTs themselves should be treated as opaque tokens.
+Clients should treat the tokens as opaque string tokens: the JWT fields and semantics are not a stable part of the specification.
+
+Servers (eg, PDS implementations) which generate and valiate auth JWTs should implement domain separation between access and refresh tokens, using the `typ` header field: access tokens should use `at+jwt`, and refresh tokens should use `refresh+jwt`.
```
Comment (devinivy) [inline src/app/[locale]/specs/xrpc/page.mdx:58]: Just a note that `at+jwt` originates from [RFC 9068
](https://www.rfc-editor.org/rfc/rfc9068.html).  I mention it because I think it could be a common misconception that it's something we invented, i.e. that "at" comes from "AT" Protocol.

src/app/[locale]/specs/xrpc/page.mdx
```diff
@@ -87,13 +89,17 @@ The current mechanism is to use short-lived JWTs signed by the account's atproto
...
 
-- `alg` header field: indicates the signing key type (see [Cryptography](/specs/cryptography))
+- `alg` header field (string, required): indicates the signing key type (see [Cryptography](/specs/cryptography))
     - use `ES256K` for `k256` keys
     - use `ES256` for `p256` keys
-- `iss` body field: account DID that the request is being sent on behalf of. This may include a suffix service identifier; see below
-- `aud` body field: service DID associated with the service that the request is being sent to
-- `exp` body field: token expiration time, as a UNIX timestamp with seconds precision. Should be a short time window, as revocation is not implemented. 60 seconds is a good value.
-- JWT signature: base64url-encoded signature using the account DID's signing key
+- `typ` header field (string, required): `JWT`
```
Comment (devinivy) [inline src/app/[locale]/specs/xrpc/page.mdx:95]: I don't think we have formally intended to specify that this must be "JWT".  We intentionally do not require that it take this value when we verify the token since we hadn't explicitly specified it.  Wanted to leave the door open to a different value we might prefer more, e.g. perhaps something akin to `at+jwt`/`refresh+jwt`/`dpop+jwt`.

Comment (bnewbold) [reply src/app/[locale]/specs/xrpc/page.mdx:95]: @matthieusieben had recommended this in https://github.com/bluesky-social/atproto-website/issues/327.

I can mention the field and that it doesn't have a stable/specified value yet for this use-case?

Comment (matthieusieben) [reply src/app/[locale]/specs/xrpc/page.mdx:95]: I indeed wanted to put something there. Whether it's JWT or something more specific is a more complex topic.

Either we consider that there will only ever be only one type of JWT in Atproto (besides access & refresh tokens) and `JWT` is fine. Or we want to be future proof and we might want to do something like `service+jwt` (or `pds-2-pds+jwt`, or whatever).

The important thing here is to make sure that a JWT generated for one particular purpose cannot be used for another purpose, while still allowing them to be signed with the same key
 a dev could choose to implement a PDS from scratch using the same key everywhere
 The only control we'd have is the typ header.

src/app/[locale]/specs/xrpc/page.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/xrpc/page.mdx]: Thanks! I didn't know that. updated text to clarify

src/app/[locale]/specs/xrpc/page.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/xrpc/page.mdx]: I updated with:

> currently `JWT`, but intend to update to a more specific value

src/app/[locale]/specs/xrpc/page.mdx
```diff
@@ -87,13 +89,17 @@ The current mechanism is to use short-lived JWTs signed by the account's atproto
...
+- `iss` body field (string, required): account DID that the request is being sent on behalf of. This may include a suffix service identifier; see below
+- `aud` body field (string, required): service DID associated with the service that the request is being sent to
+- `exp` body field (number, required): token expiration time, as a UNIX timestamp with seconds precision. Should be a short time window, as revocation is not implemented. 60 seconds is a good token lifetime.
+- `iat` body field (number, required): token creation time, as a UNIX timestamp with seconds precision.
+- `lxm` body field (string, optional): short for "lexicon method". NSID syntax. Indicates the endpoint that this token authorizes. Servers must always validate this field if included, and should require it for security-sensitive operations. May become required in the future.
+- `jti` body field (string, required): unique random string nonce. May be used by receiving services to prevent reuse of token and replay attacks.
+- JWT signature (string, required): base64url-encoded signature using the account DID's signing key
```
Comment (devinivy) [inline src/app/[locale]/specs/xrpc/page.mdx:102]:
```suggestion
- JWT signature (string, required): base64url-encoded signature using the account DID's signing key.
```

src/app/[locale]/specs/xrpc/page.mdx
```diff
@@ -87,13 +89,17 @@ The current mechanism is to use short-lived JWTs signed by the account's atproto
...
 
-- `alg` header field: indicates the signing key type (see [Cryptography](/specs/cryptography))
+- `alg` header field (string, required): indicates the signing key type (see [Cryptography](/specs/cryptography))
     - use `ES256K` for `k256` keys
     - use `ES256` for `p256` keys
-- `iss` body field: account DID that the request is being sent on behalf of. This may include a suffix service identifier; see below
-- `aud` body field: service DID associated with the service that the request is being sent to
-- `exp` body field: token expiration time, as a UNIX timestamp with seconds precision. Should be a short time window, as revocation is not implemented. 60 seconds is a good value.
-- JWT signature: base64url-encoded signature using the account DID's signing key
+- `typ` header field (string, required): currently `JWT`, but intend to update to a more specific value
```
Comment (devinivy) [inline src/app/[locale]/specs/xrpc/page.mdx:95]:
```suggestion
- `typ` header field (string, required): currently `JWT`, but intend to update to a more specific value.
```

devinivy (approved):

PR #362, opened 2024-11-03
Configure MDX editor support

remcohaszing:
By default a TypeScript program includes all TypeScript files. However, because there’s an `include` pattern, `**/*.mdx`, has to be added for editor support.

The `mdx.checkMdx` option turns subtle error hints in the editor into red squiggly lines.

The global type `MDXProvidedComponents` configures the type of components injected into MDX. This provides support for `<Note />`, `<Image />`, etc.

The `useMDXComponents()` function does not take any arguments. The documentation on which this was based, is wrong.
(https://github.com/vercel/next.js/pull/67188)

For details, see https://github.com/mdx-js/mdx-analyzer

### DIFF
--- a/mdx-components.tsx
+++ b/mdx-components.tsx
@@ -1,10 +1,9 @@
-import { type MDXComponents } from 'mdx/types'
-
 import * as mdxComponents from '@/components/mdx'
 
-export function useMDXComponents(components: MDXComponents) {
-  return {
-    ...components,
-    ...mdxComponents,
-  }
+declare global {
+  type MDXProvidedComponents = typeof mdxComponents
+}
+
+export function useMDXComponents() {
+  return mdxComponents
 }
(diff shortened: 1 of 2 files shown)


gaearon:
Thanks! How do we test the change?

remcohaszing:
You’ll notice the difference in your editor. You need to use either the MDX language server or the MDX TypeScript plugin. Exact configurations can differ per editor. The [MDX VSCode extension](https://marketplace.visualstudio.com/items?itemName=unifiedjs.vscode-mdx) supports it by default. I will assume VSCode for the purpose of explaining.

Because MDX files are now included in the same TypeScript program that’s defined by `tsconfig.json`, you will now get matching IntelliSense, such as hover information, _Go to definition_ support, autocompletions, etc.

Hover before:

A tooltip showing an unresolved import.

Hover after:

A tooltip showing a resolved type.

Because of the global `MDXProvidedComponents`, you now get completions and type errors for components provided via `mdx-components.tsx`:

A JSX prop marked as an error and JSX prop completion items.

gaearon (approved):

PR #367, opened 2024-11-24
Update en.mdx

e16m:
Noticed a typo when reading the Glossary section for the first time. I believe the author meant _more_ not _move_. If that's not correct then please ignore.


bnewbold (approved):
thanks!

PR #369, opened 2024-11-27
fix: Add label to language changer

6TELOIV:
Add `aria-label="Language"` to the language changer for a11y purposes, so screen readers will announce the label of the control.

I think there's a potential case to be made for a visual label instead, maybe through an icon like this lucide languages one. But, for sighted users it's fairly obvious that it's a language selector, so it probably is not be needed.


bnewbold (approved):
thanks!

PR #374, opened 2024-11-30
specs: 'nsid' as a record key format

bnewbold:
Probably shouldn't merge until we get support in to typescript codebase.

This will be used by lexicon resolution.


dholms (approved):
looks good 👍 

I'm not sure there's any actual code changes that need to be done to support this in the TS code. I don't think we ever validate the content of the `key` field of a lexicon

bnewbold:
okey doke, will go ahead and merge!

PR #375, opened 2024-12-04
specs: Sync (including firehose)

bnewbold:
This finally puts the firehose (`com.atproto.sync.subscribeRepos` event stream) into the actual atproto specs, at least in current form. This document was originally drafted two months ago.

This has gone through some internal review and revision, but we are open to feedback on the structure and content of this spec: should some parts (backfill and validation) be broken out to "user guides" or other less formal documents? Should the specs avoid getting in to field-level specifics? `subscribeRepos` is a bit special and important, but the details are not really core or essential to the design of the protocol.

There is also a decent change that firehose and data synchronization semantics generally will change in the near future. We want to support easier backfill and filtering for subsets of the network, and may try to find ways to fit in non-archival relays as first-class network components. But also might still reverse course, remains to be seen.


src/app/[locale]/specs/sync/page.mdx
```diff
@@ -0,0 +1,171 @@
...
+
+## Firehose
+
+The repository event stream (com.atproto.sync.subscribeRepos, also called the "firehose") is an Event Stream which broadcasts updates to repositories (#commit events), handles and DID documents (#identity), and account hosting status ( #account). PDS hosts provide a single stream with updates for all locally-hosted accounts. "Relays" are network services which subscribe to one or more repo streams (eg, multiple PDS instances) and aggregate them in to a single combined repo stream. The combined stream has the same structure and event types. A Relay which aggregates nearly all accounts from nearly all PDS instances in the network (possibly through intermediate relays) outputs a “full-network” firehose.  Relays often mirror and can re-distribute the repository contents, though their core functionality is to verify content and output a unified firehose.
```
Comment (devinivy) [inline src/app/[locale]/specs/sync/page.mdx:19]:
```suggestion
The repository event stream (`com.atproto.sync.subscribeRepos`, also called the "firehose") is an Event Stream which broadcasts updates to repositories (`#commit` events), handles and DID documents (`#identity`), and account hosting status (`#account`). PDS hosts provide a single stream with updates for all locally-hosted accounts. "Relays" are network services which subscribe to one or more repo streams (eg, multiple PDS instances) and aggregate them in to a single combined repo stream. The combined stream has the same structure and event types. A Relay which aggregates nearly all accounts from nearly all PDS instances in the network (possibly through intermediate relays) outputs a “full-network” firehose.  Relays often mirror and can re-distribute the repository contents, though their core functionality is to verify content and output a unified firehose.
```

src/app/[locale]/specs/sync/page.mdx
```diff
@@ -0,0 +1,171 @@
...
+
+Intermediate services (eg, relays) may chose to modify or pass through identity events:
+
+- they may replace the handle with the result of their own resolution; or always remove the handle field; or always pass it through unaltered
+- they may filter out identity events if they observe that identity has not actually changed
+- they may emit identity events based on changes they became aware of independently (eg, via periodic re-validation of handles)
+
+### `#account` Events
+
+Indicates that there may have been a change in account status at the service which emits the event, and what the new status is. For example, it could be the result of creation, deletion, or temporary suspension of an account. The event describes the current status, not what changed.
```
Comment (devinivy) [inline src/app/[locale]/specs/sync/page.mdx:56]: Would it be appropriate to refer to it as not just "status" but "hosting status"?

src/app/[locale]/specs/sync/page.mdx
```diff
@@ -0,0 +1,171 @@
...
+    - `deleted`: account has been deactivated, possibly permanently.
+    - `deactivated`: temporary or indefinite removal of all public data by the account themselves.
+
+When coming from any service which redistributes account data, the event describes what the new status is *at that service*, and is authoritative in that context. In other words, the event is hop-by-hop for repository hosts and mirrors.
+
+See the Account Hosting specification for more details.
+
+### `#commit` Events
+
+This event indicates that there has been a new repository commit for the indicated account. The event usually contains the "diff" of repository data, in the form of a CAR slice.
```
Comment (devinivy) [inline src/app/[locale]/specs/sync/page.mdx:76]: We could link to the doc on CAR diffs.

src/app/[locale]/specs/sync/page.mdx
```diff
@@ -0,0 +1,171 @@
...
+- services should verify that repository data structures are valid against the specification. Missing fields, incorrect MST structure, or other protocol-layer violations should result in events being rejected.
+- services may apply rate-limits to identity, account, and commit events, and suspend accounts or upstream services which violate those limits. Rate limits might also be applied to recovery modes such as invalid signatures resulting in an identity refresh, `tooBig` events, missing or out-of-order commits, etc.
+- services should ignore commit events with a `rev` lower or equal to the most recent processed `rev` for that DID, and should reject commit events with a `rev` corresponding to a future timestamp (beyond a clock drift window of a few minutes)
+- services should check the `since` value in commit events, and if it does not match the previous seen `rev` for that DID, mark the repo as out-of-sync (similar to a `tooBig` commit event)
```
Comment (devinivy) [inline src/app/[locale]/specs/sync/page.mdx:113]: Is there a nuance here, i.e. could `since` be `<=` previous rev?  IIRC this would at least be in line with the repo diffs.  Though at the same time, we don't necessarily need to permit it here.

src/app/[locale]/specs/sync/page.mdx
```diff
@@ -0,0 +1,171 @@
...
+- data limits on records specifically should be verified. events containing corrupt or entirely invalid records may be rejected. for example, a record not being CBOR at all, or exceeding normal data size limits.
```
Comment (devinivy) [inline src/app/[locale]/specs/sync/page.mdx:114]:
```suggestion
- data limits on records specifically should be verified. Events containing corrupt or entirely invalid records may be rejected. for example, a record not being CBOR at all, or exceeding normal data size limits.
```

src/app/[locale]/specs/sync/page.mdx
```diff
@@ -0,0 +1,171 @@
...
+
+Commit events are broadcast when the account repository changes. Commits can be "empty", meaning no actual record content changed, and only the `rev` was incremented. They can contain a single record update, or multiple updates. Only the commit object, record blocks, and MST tree nodes are authenticated (signed): the `since`, `ops`, `blobs`, and `tooBig` flag are not self-certifying, and could in theory be manipulated, or otherwise be incorrect or incomplete.
+
+If `since` is not included, the commit should include the full repo tree, or set the `tooBig` flag.
+
+If the `tooBig` flag is set, then the amount of updated data was too much to be serialized in a single stream event message. Downstream services which want to maintain complete synchronized copies for the repo need to fetch the diff separately, as discussed below.
+
+### Firehose Validation Best Practices
+
+A service which does full validation of upstream events has a number of properties to track and check. For example, Relay instances should fully validation content from PDS instances before re-broadcasting.
```
Comment (saranrapjs) [inline src/app/[locale]/specs/sync/page.mdx:101]:
```suggestion
A service which does full validation of upstream events has a number of properties to track and check. For example, Relay instances should fully validate content from PDS instances before re-broadcasting.
```

ngerakines:
> ... should some parts (backfill and validation) be broken out to "user guides" or other less formal documents?

I think that is pretty reasonable.

> Should the specs avoid getting in to field-level specifics? subscribeRepos is a bit special and important, but the details are not really core or essential to the design of the protocol.

The documentation should list and be explicit with parameters, fields, and expected behavior to avoid the implication that the behavior could vary between implementations.

Everything else can and should be dumped into user guides.

src/app/[locale]/specs/sync/page.mdx
```diff
@@ -0,0 +1,171 @@
...
+
+When all of the accounts are `synchronized`, the process is complete. At large scale it may be hard to get perfect synchronization: PDS instances may be down at various times, identities may fail to resolve, or invalid events, data, or signatures may end up in the network.
+
+## Security Concerns
+
+General mitigations for resource exhaustion attacks are recommended: event rate-limits, data quotas per account, limits on data object sizes and deserialized data complexity, etc.
+
+Care should always be taken when making network requests to unknown or untrusted hosts, especially when the network locators for those host from from untrusted input. This includes validating URLs to not connect to local or internal hosts (including via HTTP redirects), avoiding SSRF in browser contexts, etc.
+
```
Comment (DavidBuchanan314) [inline src/app/[locale]/specs/sync/page.mdx:164]: Might want to draw specific attention to being mindful of *outbound* request rates.

Consider multiple PDSes colluding to spam `#identity` events, each below the inbound event rate limits for an individual PDS, but collectively driving lots of handle-resolution traffic towards hosts of their choice.

Comment (DavidBuchanan314) [reply src/app/[locale]/specs/sync/page.mdx:164]: This might even be relevant for non-malicious traffic, I vaguely recall someone talking about getting blocked by their DNS resolver for trying to resolve too many handles.

bnewbold:
@ngerakines agree on specifics! what I forgot to mention is that these are already laid out in a machine-readable format in the Lexicons. there is some tension (at least to me) around the degree to which the `com.atproto.*` lexicons are part of the protocol or not.

antstanley:
Adding more docs around `com.atproto.sync.subscribeRepo` is great. I've recently been creating a feed builder, and finding docs in one place on the firehose has been challenging. Spent too much time reading source code, and Github comments.

The docs look good, but I would add more detail around serialization/encoding methods, [DAG-CBOR](https://ipld.io/docs/codecs/known/dag-cbor/) and [CAR](https://ipld.io/specs/transport/car/). An explanation on how/where they are used, and also how frames are implemented with DAG-CBOR.

Either way I lost a day to that detail, and having it documented would be great!

bumblefudge:
> I would add more detail around serialization/encoding methods, [DAG-CBOR](https://ipld.io/docs/codecs/known/dag-cbor/) and [CAR](https://ipld.io/specs/transport/car/). An explanation on how/where they are used, and also how frames are implemented with DAG-CBOR.

@antstanley see #344 from a few months back

src/app/[locale]/specs/sync/page.mdx
```diff
@@ -0,0 +1,171 @@
...
+## Synchronization Primitives
+
+As described in the repository spec, each commit to a repository has a *revision* number, in TID syntax. The revision number for an account must always increase, even if the account is migrated between hosts, or has a period of inactivation in the network. To make this simple, it is recommended to just use the current time as the TID, even when starting with a new account. Services should reject or ignore revision numbers corresponding to future timestamps (beyond a short fuzzy time drift window). Network services can track the commit revision for every account they have seen, and use this to verify synchronization progress. Services which synchronize data can include the most-recently-processed revision in HTTP responses to API requests from the account in question, in the `Atproto-Repo-Rev` HTTP response header. This allows clients (and users) to detect if the response is up-to-date with the actual repository, and detect any problems with synchronization.
```
Comment (bumblefudge) [inline src/app/[locale]/specs/sync/page.mdx:15]:
```suggestion
As described in the repository spec, each commit to a repository has a *revision* number, in TID syntax. The revision number for successive commits must always increase, even if the account is migrated between hosts, or has a period of inactivation in the network. The revision number of the most recent commit can serve as a counter to guide synchronizations. To make this simple, it is recommended to just use the current time as the TID for each commit, even for the initial commit when creating a new account. Services should reject or ignore revision numbers corresponding to future timestamps (beyond a short fuzzy time drift window). Network services can track the commit revision for every account they have seen, and use this to verify synchronization progress. Services which synchronize data can include the most-recently-processed revision in HTTP responses to API requests from the account in question, in the `Atproto-Repo-Rev` HTTP response header. This allows clients (and users) to detect if the response is up-to-date with the actual repository, and detect any problems with synchronization.
```

Comment (bumblefudge) [reply src/app/[locale]/specs/sync/page.mdx:15]: also what's with `inactivation`? is that a deliberate portmanteau of 'inactivity' and 'deactivation'? 😉

src/app/[locale]/specs/sync/page.mdx
```diff
@@ -0,0 +1,171 @@
...
+
+Identity and account information is *not* self-certifying, and services may need need to verify independently. This usually means independent DID and handle resolution. Account hosting status might also be checked at account PDS hosts, to disambiguate hosting status at different pieces of infrastructure.
+
+The event message types are declared in the `com.atproto.sync.subscribeRepos` Lexicon schema, and are summarized below. A few fields are the same for all event types (except for `repo` vs `did` for `#commit` events):
+
+- `seq` (integer, required): used to ensure reliable consumption, as described in Event Streams
+- `did` / `repo`(string with DID syntax, required): the account/identity associated with the event
+- `time` (string with datetime syntax, required): an informal and non-authoritative estimate of when event was received. Intermediate services may decide pass-through this field as-is, or update to the current time
```
Comment (bumblefudge) [inline src/app/[locale]/specs/sync/page.mdx:29]:
```suggestion
- `time` (string with datetime syntax, required): an informal and non-authoritative estimate of when event was received. Intermediate services may decide to pass this field through to other participants as-is, or update to the current time
```

src/app/[locale]/specs/sync/page.mdx
```diff
@@ -0,0 +1,171 @@
...
+Event fields:
+
+- `seq` (integer, required): same for all event types
+- `did` (string with DID syntax, required): same for all event types
+- `time` (string with datetime syntax, required): same for all event types
+- `handle` (string with handle syntax, optional): the current handle for this identity. May be `handle.invalid` if the handle does not currently resolve correctly.
+
+Presence or absence of the `handle` field does not indicate that it is the handle which has changed.
+
+The semantics and expected behavior are that downstream services should refresh their identity data (resolve the DID and handle) for the indicated DID.
```
Comment (bumblefudge) [inline src/app/[locale]/specs/sync/page.mdx:44]:
```suggestion
The semantics and expected behavior are that downstream services should refresh their identity data from the canonical source (i.e. resolve the DID and verify the handle(s) therein) after each `#identity` event. 
```
feels like this behavioral expectation is worth being pedantically explicit about

src/app/[locale]/specs/sync/page.mdx
```diff
@@ -0,0 +1,171 @@
...
+- `seq` (integer, required): same for all event types
+- `did` (string with DID syntax, required): same for all event types
+- `time` (string with datetime syntax, required): same for all event types
+- `handle` (string with handle syntax, optional): the current handle for this identity. May be `handle.invalid` if the handle does not currently resolve correctly.
+
+Presence or absence of the `handle` field does not indicate that it is the handle which has changed.
+
+The semantics and expected behavior are that downstream services should refresh their identity data (resolve the DID and handle) for the indicated DID. 
+
+Identity events are emitted on a "best-effort" basis. Is is possible for the DID document or handle resolution status to change without any atproto service detecting the change, in which case an event would not be emitted. It is also possible for the event to be emitted redundantly, when nothing has actually changed.
```
Comment (bumblefudge) [inline src/app/[locale]/specs/sync/page.mdx:46]:
```suggestion
Identity events are emitted on a "best-effort" basis. It is possible for the DID document or handle resolution status to change without any atproto service detecting the change, in which case an event would not be emitted. It is also possible for the event to be emitted redundantly, when nothing has actually changed.
```

src/app/[locale]/specs/sync/page.mdx
```diff
@@ -0,0 +1,171 @@
...
+
+Intermediate services (eg, relays) may chose to modify or pass through identity events:
+
+- they may replace the handle with the result of their own resolution; or always remove the handle field; or always pass it through unaltered
+- they may filter out identity events if they observe that identity has not actually changed
```
Comment (bumblefudge) [inline src/app/[locale]/specs/sync/page.mdx:51]: not sure these are good ideas, haha.  just as a framing/psychological nit, it might be better to rephrase these as "things you can't assume intermediate services aren't doing to you" rather than as "things you can do to other participants as an intermediate service"?

src/app/[locale]/specs/sync/page.mdx
```diff
@@ -0,0 +1,171 @@
...
+- `since` (string with TID syntax, nullable): indicates the `rev` of a preceding commit, which the the repo diff contains differences from
+- `commit` (cid-link, required): CID of the commit object (in `blocks`)
+- `tooBig` (boolean, required): if true, indicates that the repo diff was too large, and that `blocks`, `ops`, and complete `blobs` are not all included
+- `blocks` (bytes, required): CAR "slice" for the corresponding repo diff. The commit object must always be included.
+- `ops` (array of objects, required): list of record-level operations in this commit: specific records created, updated, deleted
+- `blobs` (array of cid-link, required): set of new blobs (by CID) referenced by records in this commit
+
+Commit events are broadcast when the account repository changes. Commits can be "empty", meaning no actual record content changed, and only the `rev` was incremented. They can contain a single record update, or multiple updates. Only the commit object, record blocks, and MST tree nodes are authenticated (signed): the `since`, `ops`, `blobs`, and `tooBig` flag are not self-certifying, and could in theory be manipulated, or otherwise be incorrect or incomplete.
```
Comment (bumblefudge) [inline src/app/[locale]/specs/sync/page.mdx:93]:
```suggestion
Commit events are broadcast when the account repository changes. Commits can be "empty", meaning no actual record content changed, and only the `rev` was incremented. They can contain a single record update, or multiple updates. Only the commit object, record blocks, and MST tree nodes are authenticated (i.e., signed): the `since`, `ops`, `blobs`, and `tooBig` properties are not self-certifying, and could in theory be manipulated, or otherwise be incorrect or incomplete.
```

src/app/[locale]/specs/sync/page.mdx
```diff
@@ -0,0 +1,171 @@
...
+
+If the `tooBig` flag is set, then the amount of updated data was too much to be serialized in a single stream event message. Downstream services which want to maintain complete synchronized copies for the repo need to fetch the diff separately, as discussed below.
+
+### Firehose Validation Best Practices
+
+A service which does full validation of upstream events has a number of properties to track and check. For example, Relay instances should fully validate content from PDS instances before re-broadcasting.
+
+Here is a summary of validation rules and behaviors:
+
+- services should independently resolve identity data for each DID. They should ignore `#commit` events for accounts which do not have a functioning atproto identity (eg, lacking a signing key, or lacking a PDS service entry, or for which the DID is tombstoned)
```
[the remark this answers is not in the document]
Comment (bumblefudge) [reply src/app/[locale]/specs/sync/page.mdx:105]: Also, 
Would it be more crisp to specify how they _know_ whether a DID has been tombstoned? i.e. "for which they have received an `#account` event updating the account's `status` to `tombstoned`"? does the directory return the same thing for a never-existed and a tombstoned account, or is there a did-resolution way of founding out a previously-valid DID has been tombstoned?

src/app/[locale]/specs/sync/page.mdx
```diff
@@ -0,0 +1,171 @@
...
+
+- track the most recent commit `rev` they have successfully processed
+- keep cached identity data, and use cache expiration to ensure periodic re-validation of that data
+- track account status
+
+Identity caches should be purged any time an `#identity` event is received. Additionally, identity resolution should be refreshed if a commit signature fails to verify, in case the signing key was updated but the identity cache has not been updated yet.
+
+When `tooBig` events are emitted on the firehose, downstream services will need to fetch the diff out-of-band. This usually means an API request to the `com.atproto.sync.getRepo` endpoint on the current PDS host for the account, with the `since` field included. The `since` value should be the most recently processed `rev` value for the account, which may or may not match the `since` field in the commit event message.
+
+If a `#commit` is received with a `since` that does not match the most recently processed `rev` for the account, and is “later” (higher value) than the most recent commit `rev` the service has processed for that account, the service may need to do the same kind of out-of-band fetch as for a `tooBig` event.
```
Comment (bumblefudge) [inline src/app/[locale]/specs/sync/page.mdx:133]:
```suggestion
If a `#commit` is received with a `since` that does not match the most recently processed `rev` for the account, and is “later” (higher value) than the most recent commit `rev` the service has processed for that account, there are likely commits missing or received out of order, which can usually be remedied by the out-of-band fetch prescribed above for each `tooBig` event received.
```

src/app/[locale]/specs/sync/page.mdx
```diff
@@ -0,0 +1,171 @@
...
+- services which subscribe directly to PDS instances should keep track of which PDS is authoritative for each DID. They should remember the host each subscription (WebSocket) is connected to, and reject `#commit` events for accounts if they come from a stream which does not correspond to the current account for that DID
+- services should track account hosting status for each DID, and ignore `#commit` events for events which are not `active`
+- services should verify commit signatures for each `#commit` event, using the current identity data. If the signature initially fails to verify, the service should refresh the identity metadata in case it had recently changed. Events with conclusively invalid signatures should be rejected.
+- services should reject any event messages which exceed reasonable limits. A reasonable upper bound for producers is 5 MBytes (for any event stream message type). Commits with too much data must use the `tooBig` mechanism, though such commits should generally be avoided in the first place by breaking them up in to multiple smaller commits.
```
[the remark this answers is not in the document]
Comment (bnewbold) [reply src/app/[locale]/specs/sync/page.mdx]: also a 1,000,000 byte limit on `blocks`, specifically!

src/app/[locale]/specs/sync/page.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/sync/page.mdx:113]: this was discussed in more detail in "Reliable Synchronization". I updated the text to say "consistent with" instead of "match", and referenced that section for details.

src/app/[locale]/specs/sync/page.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/sync/page.mdx:164]: added a note

bnewbold:
@antstanley: I added links to the "Event Stream" and "Repository" specs, which talk about the wire format, CAR files, and "diffs". The "Data Model" spec talks about CBOR encoding.

You might also be interested in this project, which aligns closely with primitives used by atproto: https://dasl.ing/

src/app/[locale]/specs/sync/page.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/sync/page.mdx:15]: "inactivity"

src/app/[locale]/specs/sync/page.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/sync/page.mdx:51]: As an example, we think it is likely a best practice for relays to de-duplicate rapidly repeated and redundant identity events on the firehose.

src/app/[locale]/specs/sync/page.mdx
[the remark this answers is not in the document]
Comment (bnewbold) [reply src/app/[locale]/specs/sync/page.mdx:105]: touched on identity vs account hosting and "tombstoned" in a separate comment on the account spec PR.

for DID PLC, the full op log is available even for tombstoned DIDs, so there is a public distinction between "now gone" and "never existed". I can't remember if we currently use different HTTP status codes or error messages to distinguish this.

for did:web, I think there is no way to tell.

bnewbold:
@bumblefudge thanks for the edits! I reworded some of them manually, and didn't take all the style nits (eg, we currently use eg not e.g.).

src/app/[locale]/specs/sync/page.mdx
```diff
@@ -0,0 +1,173 @@
...
+
+## Synchronization Primitives
+
+As described in the repository spec, each commit to a repository has a *revision* number, in TID syntax. The revision number must always increase between commits for the same account, even if the account is migrated between hosts or has a period of inactivity in the network. Revision numbers can be used as a logical clock to aid synchronization of individual accounts. To keep this simple, it is recommended to use the current time as a TID for each commit, including the initial commit when creating a new account. Services should reject or ignore revision numbers corresponding to future timestamps (beyond a short fuzzy time drift window). Network services can track the commit revision for every account they have seen, and use this to verify synchronization progress. Services which synchronize data can include the most-recently-processed revision in HTTP responses to API requests from the account in question, in the `Atproto-Repo-Rev` HTTP response header. This allows clients (and users) to detect if the response is up-to-date with the actual repository, and detect any problems with synchronization.
```
Comment (snarfed) [inline src/app/[locale]/specs/sync/page.mdx:15]: I know using timestamp as rev is ubiquitous, but if it's only recommended, not required, then services probably shouldn't try to parse revs as TIDs, or at least should fail that gracefully, and definitely not reject or ignore commits based on it, right?

src/app/[locale]/specs/sync/page.mdx
```diff
@@ -0,0 +1,173 @@
...
+
+When `tooBig` events are emitted on the firehose, downstream services will need to fetch the diff out-of-band. This usually means an API request to the `com.atproto.sync.getRepo` endpoint on the current PDS host for the account, with the `since` field included. The `since` value should be the most recently processed `rev` value for the account, which may or may not match the `since` field in the commit event message.
+
+If a `#commit` is received with a `since` that does not match the most recently processed `rev` for the account, and is “later” (higher value) than the most recent commit `rev` the service has processed for that account, the service may need to do the same kind of out-of-band fetch as for a `tooBig` event.
+
+Services should keep track of the `seq` number of their upstream subscriptions. This should be stored separately per-upstream, even if there is only a single Relay connection, in case a different Relay is subscribed to in the future (which will have different `seq` numbers).
```
Comment (snarfed) [inline src/app/[locale]/specs/sync/page.mdx:135]: Is it worth saying something about seq resets here? Not sure what, but something? 🤷

src/app/[locale]/specs/sync/page.mdx
```diff
@@ -0,0 +1,173 @@
...
+
+Events can be processed concurrently, but they should be processed sequentially in-order for any given account. This can be accomplished by partitioning multiple workers using the repo DID as a partitioning key.
+
+Services can confirm that they are consuming content reliably by fetching a snapshot of repository DIDs and `rev` numbers from other services, including PDS hosts and Relay instances. After a short delay, these can be compared against the current state of the service to identify any accounts which have lower than expected `rev` numbers. These repos can then be updated out-of-band.
```
Comment (snarfed) [inline src/app/[locale]/specs/sync/page.mdx:139]: How? Would this mechanism be off-protocol? Or `listRepos`? ...which is doable, but wouldn't be a snapshot, right?

src/app/[locale]/specs/sync/page.mdx
```diff
@@ -0,0 +1,173 @@
...
+
+## Bootstrapping a Live Mirror
+
+The firehose can be used to follow new data updates, and repo exports can be used for snapshots. Actually combining the two to bootstrap a complete live-updating mirror can be a bit tricky. One approach is described below.
+
+Keep a sync status table for all accounts (DIDs) encountered. The status can be:
```
Comment (snarfed) [inline src/app/[locale]/specs/sync/page.mdx:145]: Have you all thought about standardizing or otherwise adding a per-repo status check XRPC method to relays or appviews? Maybe the biggest problem I have with the Bridgy Fed PDS right now is that accounts occasionally get "stuck" and new records aren't reflected in the appview. Committing new records occasionally fixes them, but often it doesn't, and I don't have a way to figure out what's wrong with them other than eg bothering @ericvolp12, which doesn't scale 🤷

snarfed (commented):
Exciting! Love seeing this go in, thanks for writing it @bnewbold!

PR #376, opened 2024-12-04
specs: Account Hosting

bnewbold:
Describes account hosting lifecycle, and a short description of account migration.

An earlier version of this document included additional detail on Account Migration and expected event ordering for Account Lifecycle Events. These made the spec document feel pretty long and implementation-guide-style, so we have pulled those out and put them here for now:

- https://github.com/bluesky-social/atproto/discussions/3175
- https://github.com/bluesky-social/atproto/discussions/3176

Those documents could maybe become guides or other posts on the atproto.com website. Or could be re-included here in the spec, based on feedback.


DavidBuchanan314:
Regarding account lifecycle events (which, as mentioned, are not self-certifying), I think there should be some explicit notion of "authoritative" vs "non-authoritative" sources.

The PDS host registered in a user's DID document is, implicitly, the authoritative source for events on that account, whereas a relay would be non-authoritative.

A relay should probably also only accept/forward account events from the authoritative source.

Downstream services (e.g. an AppView) have no direct way of knowing whether an event originated from a PDS (and was forwarded via a relay) or if it originated from the relay itself. These downstream services should ideally be untrusting of non-authoritative events by default, and query the PDS before taking action (depending on the event type).

## Future improvement ideas:

Querying the PDS could be avoided by having the PDS sign lifecycle events, using a key specified in the PDS's DID document. If the relay is producing a lifecycle event of its own accord (e.g. takedowns) then it could sign it with its own key.

This makes account events semi-self-certifying. The user explicitly delegates event-sourcing-authority to their PDS host, via their DID document. Downstream services like appviews can choose to trust additional authorities according to their own policy (e.g. a particular relay might be trusted to emit account takedown events)

bnewbold:
@DavidBuchanan314 an earlier version of this did actually have a whole concept of the PDS being the "authority" for overall account hosting status, with each "downstream" service then optionally layering additional state on top.

One reason we didn't stick with that is that we want to leave the door open (or reflect the reality of) downstream services just leaving this all up to the "next hop". AKA, probably don't want all downstream services constantly validating every `#account` message they see with a PDS hit (especially for big accounts, all at the same time).

In terms of making account events self-certifying, we are a bit hesitant to add signatures around some events like this. This kind of signature is not deniable (we don't have PFS), and we don't have strong timestamps (distributed system), so it opens up some weird hypotheticals around re-broadcasting events. It also maybe makes all infra takedowns more enumerable for all time, which might be good (accountability!) or bad (could leak methods, drives discourse).

Overall, I think we aren't totally against the concept of PDS being the authority for account state (and even the user, in the case of deactivation and deletion), we just are not so confident as to codify it right now.

DavidBuchanan314:
You could have a signature validity window on the order of 24h (based on whatever the signer declared the timestamp to be, at time of signing).

Since the key would be declared in a DID document, you could rotate it regularly. So long as nobody caches the DID resolution for longer than some fixed period (e.g. 24h) then after 48h (signature validity window + DID TTL) it would be safe to publish your old signing keys - thus making any prior signatures deniable.

I can understand the hesitance though, there are a lot of things to consider and I can't be confident I know what all of those things are.

src/app/[locale]/specs/account/page.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+The deactivation and suspension states are implied to be temporary, though they might have indefinite time periods. Deletion and takedowns are implied to be more final, but can still technically be reverted (and frequently are, in practice).
+
+Downstream services might decide to delete content rapidly upon takedown (like deletion), or simply stop redistributing content until a later time.
+
+Identities might also be tombstoned, which is also implied to be permanent and final, but can technically be reversed in certain conditions. If the account's identity is `tombstoned`, the account hosting status should be interpreted as `deleted`.
```
Comment (bumblefudge) [inline src/app/[locale]/specs/account/page.mdx:36]: should `tombstoned` be in the list above? or are tombstoned a subset of "status==`deleted`" accounts differentiated at some other layer?  "interpreted as" is a little confusing here, couldn't hurt to be more explicit about what's same-as and what's different, particularly if `tombstoned` is a member of the enum above.

src/app/[locale]/specs/account/page.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+By default, if an account’s current active PDS is not available, account status should remain unchanged, at least for some time period. This keeps the network resilient and accounts online during temporary infrastructure outages. Services may decide their own policies for accounts whose PDS hosts remain offline for long periods of time.
+
+The general expectation is that downstream services will not list an account as `active` if it is inactive at an upstream service. However, the concept of “upstream” is informal, and services may need to define their own policies for some corner-cases. For example, an AppView might consume from multiple independent relay providers, who report differing statuses for the same account, due to policy differences. If the network connection between a relay and PDS is disruptive, they may report different statuses for the same account.
```
Comment (bumblefudge) [inline src/app/[locale]/specs/account/page.mdx:54]:
```suggestion
The general expectation is that downstream services will not list an account as `active` if it is inactive at an upstream service. However, the concept of “upstream” is informal, and services may need to define their own policies for some corner-cases. For example, an AppView might consume from multiple independent relay providers, who report differing statuses for the same account, due to policy differences. If the network connection between a relay and PDS is disrupted, they may report different statuses for the same account. 
```

src/app/[locale]/specs/account/page.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+The general expectation is that downstream services will not list an account as `active` if it is inactive at an upstream service. However, the concept of “upstream” is informal, and services may need to define their own policies for some corner-cases. For example, an AppView might consume from multiple independent relay providers, who report differing statuses for the same account, due to policy differences. If the network connection between a relay and PDS is disruptive, they may report different statuses for the same account. 
+
+If there is doubt about about an account’s general status in the network, the account’s current active PDS host can be queried using the `com.atproto.sync.getRepoStatus` endpoint. If the account is not `active` there, it is generally expected to not be active elsewhere, especially in the case of user-initiated actions (deactivation or deletion).
```
Comment (bumblefudge) [inline src/app/[locale]/specs/account/page.mdx:56]: "generally expected" feels a little hedgey-- shouldn't it be more definitive, as in, "the PDS host currently listed in the DID doc is authoritative" or "believe a status to the contrary at your own risk" or that kind of thing?

src/app/[locale]/specs/account/page.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+### Content Deletion
+
+Separate from overall account status, individual pieces of content may be removed. When accounts remove records from their public repositories, hosting services should remove public access to that content.
```
Comment (bumblefudge) [inline src/app/[locale]/specs/account/page.mdx:60]: Would this be crisper if it explicitly said the authoritative way of knowing whether an account has "removed records from its public repos"? i.e., if you (as an intermediate service or consumer) check the current PDS host for a piece of content you got elsewhere or earlier, and current host says 404, you SHOULD flush your cache and stop distributing it to others.

src/app/[locale]/specs/account/page.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+Applications which would benefit from explicit “tombstones” (indications that content previously existed) should explicitly design them in to Lexicon schemas. In the absence of explicit application-level tombstones, services are expected not to differentiate between content which has never existed and content which has been entirely deleted. Note that this is different from acknowledging account-level deletion, which is encouraged.
```
Comment (bumblefudge) [inline src/app/[locale]/specs/account/page.mdx:62]:
```suggestion
Applications which would benefit from explicit “tombstones” (indications that content previously existed) should explicitly design them in to Lexicon schemas. In the absence of explicit application-level tombstones, services are expected not to differentiate between content which has never existed and content which has been entirely deleted. Note that this is different from acknowledging account-level deletion, which is encouraged; see the "#account Events" section of the Sync Specification for more details.
```

src/app/[locale]/specs/account/page.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+At a protocol level, this is how account migration works. And in some situations, when any previous PDS instance is unavailable or has inactive account status, getting in to the above state may be the only account recovery path available.
+
+However, in the common case, when there is an active account on a functional current PDS, a more seamless account migration process is possible. It can summarized in a few high-level steps:
+
+- creating an account on the new PDS (which may start with `active` false, such as `deactivated`)
+- migrating data from the old PDS to the new PDS
+- updating identity (DID document) to point to the new PDS and using a new atproto signing key. For PLC DIDs, usually involves updating PLC rotation keys as well.
+- updating account status on both PDS instances, such that old is inactive, and new is active
+- emitting a `#commit` event from the new PDS, signed by the new atproto signing key, with a higher commit `rev`.
```
Comment (bumblefudge) [inline src/app/[locale]/specs/account/page.mdx:76]: is there a dedicated/internal/"dummy" commit that needs to be emitted, or is this a "new skeet"/content commit? is a migration not valid/complete/official until that first commit emitted from the new PDS?

Comment (snarfed) [reply src/app/[locale]/specs/account/page.mdx:76]: I wonder if this is actually a requirement...? ie, if an account's DID doc is updated to point to a new PDS, its account is created there as active, its repo head there is the same `rev` as before, and its account is updated on the old PDS to be inactive, should other services still not recognize the migration if the repo is missing a new commit?

src/app/[locale]/specs/account/page.mdx
```diff
@@ -0,0 +1,90 @@
...
+At a high level, the current PDS host for an account is indicated in the account’s DID document. If a DID document is updated to point at a new PDS host, the account status is `active` at that host, and there is a functioning repository (with valid signature and commit `rev`), then the account has effectively been migrated.
+
+At a protocol level, this is how account migration works. And in some situations, when any previous PDS instance is unavailable or has inactive account status, getting in to the above state may be the only account recovery path available.
+
+However, in the common case, when there is an active account on a functional current PDS, a more seamless account migration process is possible. It can summarized in a few high-level steps:
+
+- creating an account on the new PDS (which may start with `active` false, such as `deactivated`)
+- migrating data from the old PDS to the new PDS
+- updating identity (DID document) to point to the new PDS and using a new atproto signing key. For PLC DIDs, usually involves updating PLC rotation keys as well.
+- updating account status on both PDS instances, such that old is inactive, and new is active
```
Comment (snarfed) [inline src/app/[locale]/specs/account/page.mdx:75]: Deactivating the account on the old PDS is recommended but not strictly required, eg for the cases when it's down or malicious, right?

src/app/[locale]/specs/account/page.mdx
```diff
@@ -0,0 +1,90 @@
...
+## PDS Account Migration
+
+At a high level, the current PDS host for an account is indicated in the account’s DID document. If a DID document is updated to point at a new PDS host, the account status is `active` at that host, and there is a functioning repository (with valid signature and commit `rev`), then the account has effectively been migrated.
+
+At a protocol level, this is how account migration works. And in some situations, when any previous PDS instance is unavailable or has inactive account status, getting in to the above state may be the only account recovery path available.
+
+However, in the common case, when there is an active account on a functional current PDS, a more seamless account migration process is possible. It can summarized in a few high-level steps:
+
+- creating an account on the new PDS (which may start with `active` false, such as `deactivated`)
+- migrating data from the old PDS to the new PDS
```
Comment (snarfed) [inline src/app/[locale]/specs/account/page.mdx:73]: It's been interesting to see the deprioritization of storing a full copy of your repo on your local devices to support migration. Totally ok! Just interesting. https://atproto.com/guides/overview#:~:text=backup

snarfed (commented):
Finally! This is great! Thanks for writing it up @bnewbold.

I left a few comments inline, nothing major. You might also consider mentioning that handle resolution and `#identity` events are out of scope here, and largely (entirely?) orthogonal to account lifecycle and PDS metadata.

Otherwise looks great!

src/app/[locale]/specs/account/page.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/account/page.mdx:36]: This gets at the distinction between "identity" and "account" in atproto, which I tried to sketch out in the intro paragraph of this doc.

An identity is a DID, and DIDs have a concept of "tombstoned". But accounts are held on a PDS instance, and those accounts don't have a lifecycle state called "tombstoned", and it is totally possible for the DID to be tombstoned and the PDS isn't aware of it (eg, a `did:web` can be updated without notifying the PDS).

This paragraph is talking about how third party services could/should reconcile possibly contradicting states between identity and account.

src/app/[locale]/specs/account/page.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/account/page.mdx:56]: We actually specifically removed "PDS is authority" language from this document (see comments with David above in this PR discussion thread).

Agree it is a bit hedge-y, but that matches the norms of these specs overall at this point. Definitely would move in explicit "MUST" / "SHOULD"  direction when we go to a standards body, but should probably revisit almost all of these details at that same time.

src/app/[locale]/specs/account/page.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/account/page.mdx:60]: The details around repositories and record deletion are mostly discussed in the Repository spec page and the Sync spec page (the later is in a concurrent PR right now). There is similar discussion in the Blobs spec page. This document ("Account Hosting") isn't about how to find out if a record has been deleted from a repo (that comes over the firehose, FWIW), it is about expectations for when that deletion event is received.

I guess it fits in a bit weird here, but it felt strange to talk so much about account-wide state changes without at least some mention of more granular content deletion at the same time.

src/app/[locale]/specs/account/page.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/account/page.mdx:76]: This section is somewhat intentionally vague. We want "new PDS" to emit some kind of commit after the migration; it doesn't really matter if it is empty or not, so we don't mention or require that. empty/no-op is easy and what we would imagine most implementations would do.

the motivation for sending a commit like this is to have all downstream services try to validate the commit signature from this PDS, with a new rev. in theory this is redundant: if they got the identity event their caches are already updated, and if everything was smooth the next commit will come through fine. in practice, there are lots of things that might be slow on the first such commit, or may not have made the leap. we think sending a first commit event immediately after migration makes it more likely that actual initial interactions (post, like, whatever) will be successfully processed.

there is not practical way to require this: implementations might just not do it, and things would "probably" be fine? but it is valuable enough to mention and encourage here.

src/app/[locale]/specs/account/page.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/account/page.mdx:75]: correct; there also isn't really a mechanism to "enforce" it being required.

it is good to do because it makes it less likely for old PDS to emit `#commit` events by accident (eg, a cronjob or integration might have valid auth against old PDS and result in repo writes, which basically fork repo state and makes it critical for relays et al to have tracked account state correctly). it also resolves issues with multiple apps/devices and user confusion; and makes it more likely for the old PDS to be able to "clean up". We could have recommended "delete" old PDS account, but that felt risks (eg, it purges repo CAR and blobs, feels good to keep those copies around at least for a while for possible recovery).

src/app/[locale]/specs/account/page.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/account/page.mdx:73]: if both PDS instances are online, then we'd probably always recommend PDS-to-PDS migration, it will just be faster and use less bandwidth/battery on a phone (for example).

we do want to have repos backed up more places under user control! in a many-app world, i'm not sure having individual apps handle this makes the most sense though. today it probably does make sense for bsky app to do this, because it is so popular. but should whtwnd, smokesignals, etc, all do full local on-device CAR file backups? maybe a dedicated mobile app for "atproto account control and backup" makes more sense. could also PLC identity key management, and maybe even E2EE DMs in the future; the later two are security sensitive and might benefit from being outside something like the Bluesky Social app (which has large feature surface).

src/app/[locale]/specs/account/page.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+Applications which would benefit from explicit “tombstones” (indications that content previously existed) should explicitly design them in to Lexicon schemas. In the absence of explicit application-level tombstones, services are expected not to differentiate between content which has never existed and content which has been entirely deleted. Note that this is different from acknowledging account-level deletion, which is encouraged.
+
+## PDS Account Migration
+
+At a high level, the current PDS host for an account is indicated in the account’s DID document. If a DID document is updated to point at a new PDS host, the account status is `active` at that host, and there is a functioning repository (with valid signature and commit `rev`), then the account has effectively been migrated.
+
+At a protocol level, this is how account migration works. And in some situations, when any previous PDS instance is unavailable or has inactive account status, getting in to the above state may be the only account recovery path available.
```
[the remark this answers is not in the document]
Comment (snarfed) [reply src/app/[locale]/specs/account/page.mdx:68]: https://github.com/bluesky-social/atproto-website/pull/376#discussion_r1870365238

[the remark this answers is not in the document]
Comment (ebwinters) [reply src/app/[locale]/specs/account/page.mdx:68]: What I'm missing is how to get the users repository data from the old PDS. Emitting a new commit makes sense to me but what about the rest of their likes etc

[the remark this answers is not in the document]
Comment (snarfed) [reply src/app/[locale]/specs/account/page.mdx:68]: Right, that's currently an open question, @bnewbold addressed it in https://github.com/bluesky-social/atproto-website/pull/376#discussion_r1870498526 . You'd need the full repo anyway before you could generate and emit a new commit.

PR #279, opened 2024-01-30
Add data validation spec

emilyliu7321:
PR content from @bnewbold

PR #378, opened 2024-12-05
data validation as a "guide"

bnewbold:
This tries to finally get https://github.com/bluesky-social/atproto-website/pull/279 merged in, as a "guide" separate from the specs themselves (in the style of https://github.com/bluesky-social/atproto-website/pull/377).

I think the text here was reviewed a long time ago, but another set of eyes would not be bad.


src/app/[locale]/guides/data-validation/en.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+
+## Recommended Data Limits
+
+The following are informal guidelines, mostly to communicate order-of-magnitude expectations, and are likely to evolve over time. They are not part of the atproto specification, but if you are pushing the limits here, you may limit interoperability.
+
+**CBOR Record Size:** try to keep individual records to a few dozen KBytes. If you need to store more data, even text data, consider using a blob instead. A reasonable maximum record size limit (`MAX_CBOR_RECORD_SIZE`) is 1 MByte.
+
+Note that event stream (firehose) "frames" may consist of multiple records, and larger limits are recommended for CBOR parsing in general, on the order of 4-5 MBytes.
```
Comment (devinivy) [inline src/app/[locale]/guides/data-validation/en.mdx:22]: I assume we're referencing the `sync.subscribeRepos` endpoint here.  There is a cap of 1mb in the `blocks` field, so indeed creation of a 1mb record would end-up being marked as `tooBig`.

src/app/[locale]/guides/data-validation/en.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+**JSON Record Size:** the CBOR encoding is "canonical" for records, so focusing only on that encoding would make sense. But sometimes it is good to also have a limit on JSON encoding size. A reasonable limit (`MAX_JSON_RECORD_SIZE`) is 2 MByte.
+
+**General string length:** an overall length limit on strings within a record, including both those with and without Lexicon-specified string lengths. Measured as bytes (UTF-8 encoded). Should try to keep these to tens of KBytes at most. For an upper-bound limit (`MAX_RECORD_STRING_LEN`), reasonable to just rely on the overall CBOR record size limit. Notably, some early implementations had a 8 KByte (8192 bytes) limit.
+
+**General `bytes` length:** same as the string limit, but for binary data (`MAX_RECORD_BYTES_LEN`). Recommend relying on the overall CBOR record size limit.
```
Comment (devinivy) [inline src/app/[locale]/guides/data-validation/en.mdx:28]: For use cases requiring more bytes, could spill over into a linked blob.

src/app/[locale]/guides/data-validation/en.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+**CID binary encoding size:** recommend an overall limit (`MAX_CID_BYTES`) of 100 bytes.
```
Comment (devinivy) [inline src/app/[locale]/guides/data-validation/en.mdx:30]: Do we support CID encodings that could end-up this large?  Or is this more general/forward-looking?

src/app/[locale]/guides/data-validation/en.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+**Container nesting depth:** for example, how many layers of map inside an array inside an array, etc. If your CBOR or JSON parsing library supports a limit, the default is probably fine. A reasonable limit (`MAX_CBOR_NESTED_LEVELS`) is 32 levels of nesting.
```
Comment (devinivy) [inline src/app/[locale]/guides/data-validation/en.mdx:32]: I don't think there's necessarily an issue here, but struck me as a little conservative.

src/app/[locale]/guides/data-validation/en.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+**Container element count:** for example, how many keys in a map, or elements in an array. If your CBOR or JSON parsing library supports a limit, the default is probably fine. A reasonable limit (`MAX_CBOR_CONTAINER_LEN`) is 128 x 1024 = 131,072 elements.
+
+**Object key string length:** for example, how many bytes (UTF-8 encoded) are allowed in any key of an object. If your CBOR or JSON parsing library supports a limit, the default is probably fine. A reasonable limit (`MAX_CBOR_OBJECT_KEY_LEN`) is 8 KBYte (8192).
```
Comment (devinivy) [inline src/app/[locale]/guides/data-validation/en.mdx:36]:
```suggestion
**Object key string length:** for example, how many bytes (UTF-8 encoded) are allowed in any key of an object. If your CBOR or JSON parsing library supports a limit, the default is probably fine. A reasonable limit (`MAX_CBOR_OBJECT_KEY_LEN`) is 8 KByte (8192).
```

src/app/[locale]/guides/data-validation/en.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+**Integers:** as mentioned [in the atproto specification](https://atproto.com/specs/data-model#data-types), it is a strongly recommended best practice to keep integer values "64-bit float safe", meaning restricting them to 53 bits of precision. This ensures compatibility with Javascript without loss of numeric precision. The specific values are `MAX_SAFE_INTEGER: 9007199254740991` and `MIN_SAFE_INTEGER: -9007199254740991`.
```
Comment (devinivy) [inline src/app/[locale]/guides/data-validation/en.mdx:38]:
```suggestion
**Integers:** as mentioned [in the atproto specification](https://atproto.com/specs/data-model#data-types), it is a strongly recommended best practice to keep integer values "64-bit float safe", meaning restricting them to 53 bits of precision. This ensures compatibility with JavaScript without loss of numeric precision. The specific values are `MAX_SAFE_INTEGER: 9007199254740991` and `MIN_SAFE_INTEGER: -9007199254740991`.
```

Comment (devinivy) [inline src/app/[locale]/guides/data-validation/en.mdx:38]: JS is mentioned here, but I believe it would apply to other common languages too like Python or C.

src/app/[locale]/guides/data-validation/en.mdx
```diff
@@ -0,0 +1,54 @@
...
+## Validation Without Schema
+
+The [atproto Lexicon system](https://atproto.com/specs/lexicon) describes a data model and schema language for validating data against a known schema. But what if you are processing data where the schema isn't known or present?
+
+Here are some guidelines, by data type:
+
+* `integer`: should have values within safe limits (discussed in limits section)
+* `string`: must be valid UTF-8 encoding. No particular Unicode normalization is expected or required. Empty strings are allowed, but it is preferred to take advantage of nullable or optional fields if possible.
+* `bytes`: may be empty (length 0)
+* `cid-link`: as discussed in the protocol specification: must be CIDv1; multibase should be raw in CBOR encoding (type `0x00`); multicodec should be `dag-cbor` (`0x71`) or `raw` (`0x55`); and multihash type SHA-256 is encouraged (but not strictly required)
```
Comment (devinivy) [inline src/app/[locale]/guides/data-validation/en.mdx:50]: (Just consistency with the rest of the list below.)
```suggestion
- `integer`: should have values within safe limits (discussed in limits section)
- `string`: must be valid UTF-8 encoding. No particular Unicode normalization is expected or required. Empty strings are allowed, but it is preferred to take advantage of nullable or optional fields if possible.
- `bytes`: may be empty (length 0)
- `cid-link`: as discussed in the protocol specification: must be CIDv1; multibase should be raw in CBOR encoding (type `0x00`); multicodec should be `dag-cbor` (`0x71`) or `raw` (`0x55`); and multihash type SHA-256 is encouraged (but not strictly required)
```

src/app/[locale]/guides/data-validation/en.mdx
```diff
@@ -0,0 +1,54 @@
...
+    'Expectations around recommended data limits and validation',
+}
+
+# Data Validation
+
+Some software does not need to worry much about Bluesky or atproto schema validation. For example, client apps and bots can generally assume that data they receive from the Bluesky API is valid, and when they try to create records the PDS is responsible for double-checking schema validation.
+
+Software which consumes directly from the event stream (firehose) should be more careful. And developers designing new tools or clever hacks should be aware of some expectations and hard limits around data validation.
+
+If an individual record fails to validate for any reason, the entire record should be ignored, but other records from the same repository should be processed. If there is a problem with the repository commit data or repository structure ("MST"), the repository should be marked as invalid, but old content does not need to be de-indexed or purged. As soon as a new valid commit is received, the repository should be marked valid again. This logic mostly relevant to feed generators, AppViews, and moderation services, all of which may be consuming from the firehose.
```
Comment (dholms) [inline src/app/[locale]/guides/data-validation/en.mdx:13]:
```suggestion
If an individual record fails to validate for any reason, the entire record should be ignored, but other records from the same repository should be processed. If there is a problem with the repository commit data or repository structure ("MST"), the repository should be marked as invalid, but old content does not need to be de-indexed or purged. As soon as a new valid commit is received, the repository should be marked valid again. This logic is mostly relevant to feed generators, AppViews, and moderation services, all of which may be consuming from the firehose.
```

src/app/[locale]/guides/data-validation/en.mdx
Comment (dholms) [reply src/app/[locale]/guides/data-validation/en.mdx:22]: hmm yeah it may be worth bumping that?

src/app/[locale]/guides/data-validation/en.mdx
Comment (dholms) [reply src/app/[locale]/guides/data-validation/en.mdx:32]: Write now when importing a repo and searching through records for blobrefs, we are willing to go 32 layers deep. Mostly noting that this limit has actual consequences right now & I tend to think 32 is a pretty nice middleground

src/app/[locale]/guides/data-validation/en.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+* `integer`: should have values within safe limits (discussed in limits section)
+* `string`: must be valid UTF-8 encoding. No particular Unicode normalization is expected or required. Empty strings are allowed, but it is preferred to take advantage of nullable or optional fields if possible.
+* `bytes`: may be empty (length 0)
+* `cid-link`: as discussed in the protocol specification: must be CIDv1; multibase should be raw in CBOR encoding (type `0x00`); multicodec should be `dag-cbor` (`0x71`) or `raw` (`0x55`); and multihash type SHA-256 is encouraged (but not strictly required)
+- `array`: may have elements of heterogeneous type, because of the flexibility Lexicon unions provide.
+- `object`: key should all be strings, and follow similar data requirements to `string` fields. Empty key strings are not allowed. Key names starting with `$` are reserved for protocol use (for example, `$bytes` and `$type`)
+- `blob`: the `ref` must be a valid `cid-link`, and have `raw` multicodec (not `dag-cbor`). `size` may not be negative. `mimeType` can not be an empty string. The "legacy" blob format should be supported for reading, but new records created must be the regular blob format.
```
Comment (dholms) [inline src/app/[locale]/guides/data-validation/en.mdx:53]: worth noting that we may eventually support other codecs than `raw`. IPFS for instance chunks large files by putting them into a merkle-dag where the leaves are `raw` but the dag nodes are `dag-cbor`. alternately, i think there is some interest in leaving the codec `raw` but using `blake3` as a hash

src/app/[locale]/guides/data-validation/en.mdx
Comment (WaveringAna) [reply src/app/[locale]/guides/data-validation/en.mdx:22]: please bump it 1mb is atrocious :p

src/app/[locale]/guides/data-validation/en.mdx
Comment (bnewbold) [reply src/app/[locale]/guides/data-validation/en.mdx:30]: This is more general/forward looking. Eg, in some contexts unsupported CIDs should be rejected, but in most places generic atproto data processing should be flexible about CIDs, to allow for evolution. But generic data libraries might want to have *some* overall limit on size, an that is what this is recommending.

src/app/[locale]/guides/data-validation/en.mdx
Comment (bnewbold) [reply src/app/[locale]/guides/data-validation/en.mdx:32]: @devinivy do you have a counter-proposal number to use?

src/app/[locale]/guides/data-validation/en.mdx
Comment (bnewbold) [reply src/app/[locale]/guides/data-validation/en.mdx]: That is true if you use floats in those languages, but if you de-serialized from JSON directly in an integer format in those languages, there would not be a 53 bit limit.

And least in Python 3, large integers round-trip in a way that large floats don't:

```
import json

int(json.loads(json.dumps({"int": 900719925474099100001}))["int"])
# GOOD: 900719925474099100001

int(json.loads(json.dumps({"int": 900719925474099100001.0}))["int"])
# BAD: 900719925474099068928
```

src/app/[locale]/guides/data-validation/en.mdx
Comment (bnewbold) [reply src/app/[locale]/guides/data-validation/en.mdx:22]: The current subscribeRepos blocks byte limit is 2,000,000 (power of 10, not mebibytes). I added a comment about this here.

src/app/[locale]/guides/data-validation/en.mdx
Comment (bnewbold) [reply src/app/[locale]/guides/data-validation/en.mdx]: added another note about that here

src/app/[locale]/guides/data-validation/en.mdx
Comment (bnewbold) [reply src/app/[locale]/guides/data-validation/en.mdx:32]: I bumped to 64 for now

src/app/[locale]/guides/data-validation/en.mdx
Comment (bnewbold) [reply src/app/[locale]/guides/data-validation/en.mdx]: Coming back around to this: I agree this will be the natural extension point (non-`raw` or adding things like blake3). I'm not sure what I think validation code should do *today* though. If we (and others) are permissive at the PDS, appview, etc, then folks will start using other CID types, and things will sort of work. But other tools like automod scanning will not work.

I kind of feel like making this more flexible will need to be an increment of the repo structure version in `commit` objects, to signify that the data model has changed.

This also potentially comes up around strict DASL conformance/validation (eg, https://ipfsfoundation.org/hypha-dasl-the-test-suite/) and how we will word/phrase things in IETF draft text.

bnewbold:
Hit review comments. Also clarified use of MiBytes (power of 2) not power of 10 for "megabytes".

src/app/[locale]/guides/data-validation/en.mdx
Comment (bnewbold) [reply src/app/[locale]/guides/data-validation/en.mdx:32]: Maybe 32 was actually motivated and not just a random number from a hat? I don't really remember. Here is a new dag-cbor implementation using 32 as a default limit:

https://pkg.go.dev/github.com/hyphacoop/go-dasl@v0.1.1/drisl#DecOptions

src/app/[locale]/guides/data-validation/en.mdx
```diff
@@ -0,0 +1,54 @@
...
+The [atproto Lexicon system](https://atproto.com/specs/lexicon) describes a data model and schema language for validating data against a known schema. But what if you are processing data where the schema isn't known or present?
+
+Here are some guidelines, by data type:
+
+- `integer`: should have values within safe limits (discussed in limits section)
+- `string`: must be valid UTF-8 encoding. No particular Unicode normalization is expected or required. Empty strings are allowed, but it is preferred to take advantage of nullable or optional fields if possible.
+- `bytes`: may be empty (length 0)
+- `cid-link`: as discussed in the protocol specification: must be CIDv1; multibase should be raw in CBOR encoding (type `0x00`); multicodec should be `dag-cbor` (`0x71`) or `raw` (`0x55`); and multihash type SHA-256 is encouraged (but not strictly required)
+- `array`: may have elements of heterogeneous type, because of the flexibility Lexicon unions provide.
+- `object`: key should all be strings, and follow similar data requirements to `string` fields. Empty key strings are not allowed. Key names starting with `$` are reserved for protocol use (for example, `$bytes` and `$type`)
```
Comment (DavidBuchanan314) [inline src/app/[locale]/guides/data-validation/en.mdx]: > Empty key strings are not allowed

Where does this constraint come from? Is it in the spec?

Comment (DavidBuchanan314) [reply src/app/[locale]/guides/data-validation/en.mdx]: "key should all be strings" -> "key must all be strings" (required by dag-cbor)

src/app/[locale]/guides/data-validation/en.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+Software which consumes directly from the event stream (firehose) should be more careful. And developers designing new tools or clever hacks should be aware of some expectations and hard limits around data validation.
+
+If an individual record fails to validate for any reason, the entire record should be ignored, but other records from the same repository should be processed. If there is a problem with the repository commit data or repository structure ("MST"), the repository should be marked as invalid, but old content does not need to be de-indexed or purged. As soon as a new valid commit is received, the repository should be marked valid again. This logic is mostly relevant to feed generators, AppViews, and moderation services, all of which may be consuming from the firehose.
+
+
+## Recommended Data Limits
+
+The following are informal guidelines, mostly to communicate order-of-magnitude expectations, and are likely to evolve over time. They are not part of the atproto specification, but if you are pushing the limits here, you may limit interoperability.
```
Comment (dholms) [inline src/app/[locale]/guides/data-validation/en.mdx]: using `limit` twice is a bit awkward

```suggestion
The following are informal guidelines, mostly to communicate order-of-magnitude expectations, and are likely to evolve over time. They are not part of the atproto specification, but if you are pushing the limits here, you may encounter interoperability issues now or in the future.
```

src/app/[locale]/guides/data-validation/en.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+**CBOR Record Size:** try to keep individual records to a few dozen KBytes. If you need to store more data, even text data, consider using a blob instead. A reasonable maximum record size limit (`MAX_CBOR_RECORD_SIZE`) is 1 MiByte (1 mebibyte, or 1,048,576 bytes). Note that as of August 2025, the `subscribeRepos` Lexicon limits `#commit` message block size to 2,000,000 bytes (not mebibytes), so only a single 1 MiByte record can be created or updated per commit.
+
+Note that event stream (firehose) "frames" may consist of multiple records, and larger limits are recommended for CBOR parsing in general, on the order of 4-5 MBytes.
+
+**JSON Record Size:** the CBOR encoding is "canonical" for records, so focusing only on that encoding would make sense. But sometimes it is good to also have a limit on JSON encoding size. A reasonable limit (`MAX_JSON_RECORD_SIZE`) is 2 MiByte (2,097,152 bytes).
```
Comment (dholms) [inline src/app/[locale]/guides/data-validation/en.mdx:24]: It's awkward that this is larger than or recommendation for event size. Any reason not to lower this to 1 MiByte?

src/app/[locale]/guides/data-validation/en.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+**General string length:** an overall length limit on strings within a record, including both those with and without Lexicon-specified string lengths. Measured as bytes (UTF-8 encoded). Should try to keep these to tens of KBytes at most. For an upper-bound limit (`MAX_RECORD_STRING_LEN`), reasonable to just rely on the overall CBOR record size limit. Notably, some early implementations had a 8 KByte (8192 bytes) limit.
+
+**General `bytes` length:** same as the string limit, but for binary data (`MAX_RECORD_BYTES_LEN`). Recommend relying on the overall CBOR record size limit. As with record size overall, if more than a few dozen KBytes are needed, recommendation is to use blobs.
+
+**CID binary encoding size:** recommend an overall limit (`MAX_CID_BYTES`) of 100 bytes.
+
+**Container nesting depth:** for example, how many layers of map inside an array inside an array, etc. If your CBOR or JSON parsing library supports a limit, the default is probably fine. A reasonable limit (`MAX_CBOR_NESTED_LEVELS`) is 64 levels of nesting.
```
Comment (dholms) [inline src/app/[locale]/guides/data-validation/en.mdx]: worth noting that we only look 32 layers deep for blob refs - i'm actually inclined to recommend a lower limit on this as services may want to scan through records looking for blobs/backlinks/etc and limiting recursion can help

honestly I have a hard time imaginging a need for anything nested even 32 deep

src/app/[locale]/guides/data-validation/en.mdx
Comment (dholms) [reply src/app/[locale]/guides/data-validation/en.mdx]: I agree just leave this limited, maybe add a note about this being a possible future extension point

src/app/[locale]/guides/data-validation/en.mdx
Comment (DavidBuchanan314) [reply src/app/[locale]/guides/data-validation/en.mdx]: One "use case" is that `getPostThread` responses may be deeply nested (each reply is a new nesting level), so if someone tried to enforce these limits on xrpc responses they could get breakage.

(p.s. millipds can process arbitrarily deep blob refs in records, but that's just a party trick really)

src/app/[locale]/guides/data-validation/en.mdx
Comment (bnewbold) [reply src/app/[locale]/guides/data-validation/en.mdx]: Added a note

src/app/[locale]/guides/data-validation/en.mdx
Comment (bnewbold) [reply src/app/[locale]/guides/data-validation/en.mdx]: Updated to "must all be strings".

I don't remember if we formally disallow empty string keys anywhere in specs or upstream, though I think we should (they are a common footgun and data storage problem, in my experience).

src/app/[locale]/guides/data-validation/en.mdx
Comment (bnewbold) [reply src/app/[locale]/guides/data-validation/en.mdx:24]: my assumption is that the same data is larger when encoded as JSON compared to CBOR. If we have a 1 MiByte CBOR limit, I think we need something a bit larger for JSON. I think it is helpful to have some limit for code that works only with JSON, but maybe this part is just confusing?

src/app/[locale]/guides/data-validation/en.mdx
Comment (bnewbold) [reply src/app/[locale]/guides/data-validation/en.mdx]: `getPostThread` API responses are JSON not CBOR though, right?

I reverted back to 32 levels.

bnewbold:
There is still a bit of discussion happening, but I made another round of updates and going to merge this now. Happy to do updates either to this guide or the specs as follow-ups.

PR #381, opened 2024-12-05
fix: raw codec is 0x55

aschmahmann:
The raw codec is 0x55 rather than 0x00 (see https://github.com/multiformats/multicodec/blob/d08f5064ce50e9e7fabc5bb9a7eeb169ccefebda/table.csv#L41).

Also see https://cid.ipfs.tech/#bafkreibjfgx2gprinfvicegelk5kosd6y2frmqpqzwqkg7usac74l3t2v4


bnewbold (approved):
Indeed! Had it right elsewhere (eg, in `data-model`) but wrong here, thanks for the catch!

PR #385, opened 2024-12-15
feat: use rehype-github-alerts for github style tip blockquote

shuuji3:
The ["Self-hosting" page](https://atproto.com/guides/self-hosting#verifying-that-your-pds-is-online-and-accessible) has a GitHub style tip blockquote but it's not on the website, though it's [rendered correctly on GitHub](https://github.com/bluesky-social/atproto-website/blob/e5fca05e94d5b07519f1cfbc4316b3cff83bfe7f/src/app/%5Blocale%5D/guides/self-hosting/en.mdx#verifying-that-your-pds-is-online-and-accessible).

To render the tip blockquote correctly, this PR used https://github.com/chrisweb/rehype-github-alerts.

### Before
Screenshot from 2024-12-16 01-50-01

### After
Screenshot from 2024-12-16 01-49-40

### On GitHub
Screenshot from 2024-12-16 01-51-15

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,6 +35,7 @@
         "react": "^18.2.0",
         "react-dom": "^18.2.0",
         "react-highlight-words": "^0.20.0",
+        "rehype-github-alerts": "3.0.0",
         "remark": "^15.0.1",
         "remark-gfm": "^4.0.0",
         "remark-mdx": "^3.0.0",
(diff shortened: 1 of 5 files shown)


bnewbold:
Thanks for pointing out this issue!

Instead of pulling a new dependency, I changed these to use the `<Note>` pattern we use elsewhere in this repo.

shuuji3:
Thanks for the update 🙂 Yes, that sounds like a better approach than introducing additional dependency. Thanks!

PR #387, opened 2024-12-25
Fix typo: firehose.ts -> ingester.ts

abraj:
## Fix typo
Fix typo corresponding to an earlier filename change: `firehose.ts` -> `ingester.ts`


bnewbold (approved):
thanks!

PR #389, opened 2025-01-05
IME scopes_supported is an array

bmillwood:
Something I ran into while trying to implement OAuth login myself.


bnewbold (approved):
I think you are correct!

matthieusieben (approved):

PR #393, opened 2025-01-25
Add docs for running example app with Jetstream

ebwinters:
Goes with https://github.com/bluesky-social/statusphere-example-app/pull/25

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -608,6 +608,126 @@
         "node": ">= 10"
       }
     },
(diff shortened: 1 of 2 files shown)


knotbin:
the fact that this has sat open for so long unmerged and unreviewed. I feel like the bluesky team is just spitting in dev's faces with documentation as shitty as it is and basically no indication of any change anytime soon.

This being in the docs would have helped me immensely. Too bad it'll never be merged in.

PR #394, opened 2025-01-29
fix link to auth

yamarten:
Because the link is invalid due to the change by #343, the path is updated from `./auth` to `./oauth`.

This is related to another potential issue. #343 adds a redirect, and it [works](https://atproto.com/specs/auth) well on the English page, but the path does not match in other languages, so it [displays](https://atproto.com/ja/specs/auth) 404. More fundamental corrections may be needed, but in any case, we will solve the impending problems.


yamarten:
Not only the link but also the wording may need to be corrected, but I hesitate to edit the specification text in my poor English.

bnewbold (approved):
thanks!

PR #399, opened 2025-02-24
lexicon resolution spec

bnewbold:
See longer design discussion: https://github.com/bluesky-social/atproto/discussions/3074

Thought about putting some or all of this under the "NSID" specification, but it is really about Lexicons, not just identifiers, so sticking it here.

Note that this also pulls out mention of `revision` in "Lexicon Files" as well. We can always put them back in with same name/type, but I think it is confusing to have that in there with no clear semantics.


src/app/[locale]/specs/lexicon/en.mdx
Comment (Hoid) [inline src/app/[locale]/specs/lexicon/en.mdx:1]: "are to* be"?

Comment (Hoid) [reply src/app/[locale]/specs/lexicon/en.mdx:1]: this is about line 393, not sure why it didn't add the comment to that line 😂

Comment (bnewbold) [reply src/app/[locale]/specs/lexicon/en.mdx:1]: "are published"; thanks for the catch!

brianolson (approved):

dholms (approved):
looks great ✅

PR #404, opened 2025-03-28
Fix links to w3c's did GH pages site

encarbassotnopot:
W3c's _did-core_ repo has been renamed to just _did_. This can be checked by visiting https://github.com/w3c/did-core, which redirects to the renamed repo.

There was a link in the documentation pointing to the old GH pages site, and I've updated it to the new one.


bnewbold (approved):
thanks!

PR #408, opened 2025-04-25
Allow direct linking to h3 headers

essential-randomness:
I often find myself wanting to link to `h3` headers in documentation (like the various subtypes of events [in the firehose spec](https://atproto.com/specs/sync#firehose)), then realize this is only supported with `h2` headers .

This PR also adds  ids + anchor links to `h3` headers. 

I considered doing this for `h4`s, but that would mean changing the `Heading` component, and I wanted to keep this review as simple as possible. The only `h4`s that currently exist are in the Korean translation, and they have an `h3` header very close (and probably shouldn't be headers at all).

### DIFF
--- a/src/components/mdx.tsx
+++ b/src/components/mdx.tsx
@@ -35,6 +35,12 @@ export const h2 = function H2(
   return <Heading level={2} {...props} />
 }
 
+export const h3 = function H3(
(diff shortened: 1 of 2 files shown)


bnewbold (approved):
this is a huge improvement for me personally, thanks!

PR #426, opened 2025-06-21
Lexicon Enhancement: Support for arrays of strings in `procedure` and `query` encodings

mikekasprzak:
TypeScript patch is here: https://github.com/bluesky-social/atproto/pull/3986

Go patch is here: ???

---

Thus far my work on adding native ActivityPub support to the Bluesky PDS hasn't required any specification changes, but this is the first.

In short, ActivityPub has [two official](https://www.w3.org/TR/activitypub/#retrieving-objects) and one unofficially supported `Content-Type` header.

* `application/ld+json; profile="https://www.w3.org/ns/activitystreams"`
* `application/activity+json`
* `application/json`

It's **not possible** to correctly interoperate with other ActivityPub servers _without_ allowing multiple input `Content-Type`'s. The document bodies are identical (all based on JSON LD), which is what I'm proposing here.

Below is an example Lexicon document demonstrating what it would look like.

```json
{
  "lexicon": 1,
  "id": "org.w3.activitypub.putInbox",
  "defs": {
    "main": {
      "type": "procedure",
      "input": {
        "encoding": [
          "application/json",
          "application/activity+json",
          "application/ld+json; profile=\"https://www.w3.org/ns/activitystreams\""
        ],
        "schema": { "type": "ref", "ref": "org.w3.activitypub.object" }
      },
      "output": {
        "encoding": "application/activity+json",
        "schema": { "type": "ref", "ref": "org.w3.activitypub.object" }
      }
    }
  }
}
```

Implementations should only ever allow 1 of the listed types.

### DIFF
--- a/src/app/[locale]/specs/lexicon/en.mdx
+++ b/src/app/[locale]/specs/lexicon/en.mdx
@@ -80,7 +80,7 @@ Type-specific fields:
 - `parameters` (object, optional): a schema definition with type `params`, describing the HTTP query parameters for this endpoint
 - `output` (object, optional): describes the HTTP response body
     - `description` (string, optional): short description
-    - `encoding` (string, required): MIME type for body contents. Use `application/json` for JSON responses.
+    - `encoding` (string | array of strings, required): MIME type for body contents. Use `application/json` for JSON responses.
     - `schema` (object, optional): schema definition, either an `object`, a `ref`, or a `union` of refs. Used to describe JSON encoded responses, though schema is optional even for JSON responses.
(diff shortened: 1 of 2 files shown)


mikekasprzak:
Tangentially, I wanted to mention that changing the spec to always require an array of encodings (even if it's just one), might be a cleaner solution, as right now there doesn't seem to be any (many?) "multiple possibilities options" in the spec (i.e. String or String Array). 

Notably, getting Lexicon to support ActivityPub required some funky changes to the lex-cli tool as many values can be a string, an object, and sometimes also an array of strings or objects. 😮. Choosing Array of Strings "always" for encodings would certainly simplify things while keeping flexibility.

PR #431, opened 2025-07-30
Suggest clients use resolveIdentity instead of resolveHandle

uniphil:
The [docs for resolveHandle](https://docs.bsky.app/docs/api/com-atproto-identity-resolve-handle) say

> Does not necessarily bi-directionally verify against the the DID document.

whereas [resolveIdentity says](https://docs.bsky.app/docs/api/com-atproto-identity-resolve-identity)

> Resolves an identity (DID or Handle) to a full identity (DID document and verified handle).

Since it's critical to bi-directionally verify DID/Handle resolution, it seems safer to mention `resolveIdentity`, whose response includes the DID without omitting an important caveat.

Alternately maybe this text could mention that a network service offering `resolveHandle` can only be relied on if it's known to do bi-directional verification. (This seems unlikely if that service is an arbitrary user PDS).


mary-ext:
need https://github.com/bluesky-social/atproto/issues/3808 to be solved first I think

PR #432, opened 2025-07-30
DID spec: Make length limit more clear

pyrox0:
2 KB is translated into 2000 bytes in most calculators, but the actual limit accepted by the reference implementation(taken from https://github.com/bluesky-social/atproto/blob/82e3e7bf6bd16ddee6bf684ad10160593b1b0671/packages/did/src/did.ts#L215-L218) is 2048 characters. Therefore, this is more clear as to the exact meaning.

Note: I cannot speak or write Korean, so I did not update that document. If someone else would like to do so, i would be glad to incorporate that change into this PR.

### DIFF
--- a/src/app/[locale]/specs/did/en.mdx
+++ b/src/app/[locale]/specs/did/en.mdx
@@ -33,7 +33,7 @@ The DID Core specification constraints on DID identifier syntax, regardless of t
 - Percent-sign (`%`) is used for "percent encoding" in the identifier section, and must always be followed by two hex characters
 - Query (`?`) and fragment (`#`) sections are allowed in DID URIs, but not in DID identifiers. In the context of atproto, the query and fragment parts are not allowed.
 
-DID identifiers do not generally have a maximum length restriction, but in the context of atproto, there is an initial hard limit of 2 KB.
+DID identifiers do not generally have a maximum length restriction, but in the context of atproto, there is an initial hard limit of 2048 characters.
 
 In the context of atproto, implementations do not need to validate percent encoding. The percent symbol is allowed in DID identifier segments, but the identifier should not end in a percent symbol. A DID containing invalid percent encoding *should* fail any attempt at registration, resolution, etc.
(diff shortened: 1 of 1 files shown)


uniphil:
even though the characters are restricted to ascii, i think keeping them called "bytes" might be usefully more explicit.

+1 for writing the number as `2048`. alternate minimal spec change could be to use `KiB` suffix instead.

pyrox0:
> even though the characters are restricted to ascii, i think keeping them called "bytes" might be usefully more explicit.


That I'm not sure about. The error message(in the code I mention above) specifically says "chars", and not bytes. However, because we're in Javascript, we're abiding by Javascript's rules, and the MDN documentation for [length](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/length) says that this is the length of the string in UTF-16 code units, and therefore the number of bytes may not match the number of code units(as characters can be one or 2 code units, and each code unit is 16 bits)

Because the above code is typed with an `unknown` input, I can't say exactly how the length will be output, because if we want the length in bytes, we need to use a javascript [TextEncoder](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder) object to encode it as UTF-8, and then get the length of that output(it outputs an array of `uint8`s, and we are just getting the length of that array, which can be known accurately with its `length` property)

TL;DR you'd want to use TextEncoder anyways, as that's more accurate at giving you a single definitive number of bytes. If you want the character count, no matter the bytes per character, you want to use the string's iterator, which outputs character-by-character the entire string, and you just get the number of elements output by that iterator. The current code implementation(I believe) will actually allow strings of up to 4KiB(4096 bytes), since it is counting code units, which are 16 bits wide each(2 bytes). This is not affected by any multi-code unit characters, since it is the number of code units overall that is given by the length.

WTL;DR: Javascript is an unholy mess of horrors when it comes to string handling, beware all ye that enter here

PR #447, opened 2025-10-22
add lexinomicon to existing schema guide

axfelix:
Add content from https://github.com/bluesky-social/atproto/discussions/4245 to ATProto docs.

The resulting page is definitely too long, but I'm temporarily reserving judgment to get this out since we're working on a much more thorough IA overhaul.


src/app/[locale]/guides/lexicon/en.mdx
```diff
@@ -257,3 +257,126 @@ If a schema must change a previously-published constraint, it should be publishe
 ## Schema distribution
 
 Schemas are designed to be machine-readable and network-accessible. While it is not currently _required_ that a schema is available on the network, it is strongly advised to publish schemas so that a single canonical & authoritative representation is available to consumers of the method.
+
+&nbsp;
```
Comment (bnewbold) [inline src/app/[locale]/guides/lexicon/en.mdx]: could drop this but nbd

bnewbold:
What I was thinking was that we would plop the style guide down on it's own page, and link to it from the Lexicon spec for now. Similar to what we do with: https://atproto.com/guides/data-validation

And then work it in with more links after the IA refactor.

bnewbold:
Sorry to keep going around on this one, but could we put it under "guides" not "specs"? Mostly for the URL. I'll be linking to this almost immediately.

Other than that good to merge

bnewbold (approved):
thanks!

src/app/[locale]/guides/lexicon/en.mdx
Comment (bnewbold) [reply src/app/[locale]/guides/lexicon/en.mdx]: (this was stale feedback)

PR #453, opened 2025-10-28
specs: remove mention of NSID fragment syntax

bnewbold:
We haven't done anything with that and are not planning to; implementations should not include support for it.

Wildcard stuff is still being debated, won't remove that for now. Did fix a small typo though.

### DIFF
--- a/src/app/[locale]/specs/nsid/en.mdx
+++ b/src/app/[locale]/specs/nsid/en.mdx
@@ -50,7 +50,7 @@ A reference regex for NSID is:
 
 ### NSID Syntax Variations
 
-When referring to a group or pattern of NSIDs, a trailing ASCII star character (`*`) can be used as a "glob" character. For example, `com.atproto.*` would refer to any NSIDs under the `atproto.com` domain authority, including nested sub-domains (sub-authorities). A free-standing `*` would match all NSIDs from all authorities. Currently, there may be only a single start character; it must be the last character; and it must be at a segment boundary (no partial matching of segment names). This means the start character must be proceeded by a period, or be a bare star matching all NSIDs.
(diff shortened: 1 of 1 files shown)


src/app/[locale]/specs/nsid/en.mdx
```diff
@@ -50,9 +50,7 @@ A reference regex for NSID is:
 
 ### NSID Syntax Variations
 
-A **fragment** may be appended to an NSID in some contexts to refer to a specific sub-field within the schema. The fragment is separated from the NSID by an ASCII hash character (`#`). The fragment identifier string (after the `#`) has the same syntax restrictions as the final segment of an NSID: ASCII alphabetic, one or more characters, length restricted, etc.
```
Comment (ThisIsMissEm) [inline src/app/[locale]/specs/nsid/en.mdx:53]: This is perhaps more relevant when writing lexicons?

Comment (bnewbold) [reply src/app/[locale]/specs/nsid/en.mdx:53]: oh, that is a good point. I think of that as a "Lexicon ref" to a specific definition, and was thinking of this "NSID fragment" as arbitrary JSON Path style syntax, but they are sort of the same thing.

Comment (ThisIsMissEm) [reply src/app/[locale]/specs/nsid/en.mdx:53]: Yeah, they're same same but different slightly. It may be worth mentioning:

> In Lexicons, you may see something that looks like a NSID but is followed by a `#` and then an identifier, this is for referring to specific field within the given NSID's lexicon.

Comment (ThisIsMissEm) [reply src/app/[locale]/specs/nsid/en.mdx:53]: i.e., being explicit that these aren't NSIDs by themselves, but a Lexicon Fragment Identifier or something like that.

Comment (bnewbold) [reply src/app/[locale]/specs/nsid/en.mdx:53]: tracking issue: https://github.com/bluesky-social/atproto-website/issues/458

PR #463, opened 2025-11-14
lexicon lang spec clarifications

bnewbold:
Trying to keep this focused to just issues raised in https://github.com/bluesky-social/atproto/discussions/4343.

This spec in general could probably use a copy-edit, but I don't want to scope-creep in this PR.

### DIFF
--- a/src/app/[locale]/specs/lexicon/en.mdx
+++ b/src/app/[locale]/specs/lexicon/en.mdx
@@ -206,7 +206,7 @@ Type-specific fields:
 
 Note that unlike `object`, there is no `nullable` field on `params`.
 
-`params` can not be declared as top-level named types in a schema `defs` array. This means that they can not be reference by a `ref` or `union`.
+`params` can not be declared as a top-level named type in a schema `defs` array. This means that they can not be referenced by a `ref` or `union`.
 
 ### `token`
 
@@ -232,7 +232,7 @@ When a `ref` points to an `object` type, the type of the encoded data is unambig
 
 A `ref` can not point to a `token` type.
 
-Refs can not be declared as top-level named types in a schema `defs` array. This means that a `ref` can not point to another `ref`, nor a `union`.
+Refs can not be declared as top-level named types in a schema `defs` array. This means that a `ref` can not point to another `ref`, nor a `union`. They can reference container types that *contain* nested `ref` types, for example an `object`.
 
 ### `union`
 
@@ -251,11 +251,9 @@ A closed union with only a single type referenced is similar to `ref`. The main 
 
 An empty closed union (a union with an empty refs list and the `closed` flag set to true) is an invalid schema.
 
-The schema definitions pointed to by a `union` must have type `object` or types with a clear mapping to an object, like a `record`. All the variants must be represented by a CBOR map (or JSON Object) and must include a `$type` field indicating the variant type. Because the data must be an object, unions can not reference `token` (which would correspond to string data).
+The schema definitions pointed to by a `union` must have type `object` or `record`. All the variants must be represented by a CBOR map (or JSON Object) and must include a `$type` field indicating the variant type. A `union` can not point to types `query`, `procedure`, `subscription`, `unknown`, `blob`, `cid-link`, `array`, `params`, `token`, `ref`, `union`, or any other currently defined types.
 
-To be more complete, `union` references must point to a type `object` or `record`. They can not point to types `query`, `procedure`, `subscription`, `unknown`, `blob`, `cid-link`, `array`, `params`, `ref`, `union`, or any other currently defined types.
-
-Unions can not be declared as top-level named types in a schema `defs` array. This means that a `union` can not reference another `union`, nor a `ref`.
+Unions can not be declared as top-level named types in a schema file `defs` map. This means that a `union` can not reference another `union`, nor a `ref`.
 
 ### `unknown`
 


src/app/[locale]/specs/lexicon/en.mdx
Comment (matthieusieben) [inline src/app/[locale]/specs/lexicon/en.mdx:36]: Will we add `permission-set` and `permission` in a separate PR?

src/app/[locale]/specs/lexicon/en.mdx
```diff
@@ -206,23 +206,33 @@ Type-specific fields:
 
 Note that unlike `object`, there is no `nullable` field on `params`.
 
+`params` can not be declared as top-level named types in a schema `defs` array. This means that they can not be reference by a `ref` or `union`.
```
Comment (dholms) [inline src/app/[locale]/specs/lexicon/en.mdx]:
```suggestion
`params` can not be declared as top-level named types in a schema `defs` array. This means that they can not be referenced by a `ref` or `union`.
```

Comment (dholms) [inline src/app/[locale]/specs/lexicon/en.mdx]: also i know `params` is technically plural, but this reads weird to me and when referred to as a type, I think we should treat it has singular. for instance: "`params`  can not be declared as a top-level named type..."

src/app/[locale]/specs/lexicon/en.mdx
```diff
@@ -235,21 +245,31 @@ Unions represent that multiple possible types could be present at this location
...
+A closed union with only a single type referenced is similar to `ref`. The main difference is that the data is required to include the `$type` field.
+
+An empty closed union (a union with an empty refs list and the `closed` flag set to true) is an invalid schema.
+
+The schema definitions pointed to by a `union` must have type `object` or types with a clear mapping to an object, like a `record`. All the variants must be represented by a CBOR map (or JSON Object) and must include a `$type` field indicating the variant type. Because the data must be an object, unions can not reference `token` (which would correspond to string data).
+
+To be more complete, `union` references must point to a type `object` or `record`. They can not point to types `query`, `procedure`, `subscription`, `unknown`, `blob`, `cid-link`, `array`, `params`, `ref`, `union`, or any other currently defined types.
```
Comment (dholms) [inline src/app/[locale]/specs/lexicon/en.mdx]: maybe worth including `token` in this list as well

src/app/[locale]/specs/lexicon/en.mdx
```diff
@@ -235,21 +245,31 @@ Unions represent that multiple possible types could be present at this location
...
 
-The schema definitions pointed to by a `union` are objects or types with a clear mapping to an object, like a `record`. All the variants must be represented by a CBOR map (or JSON Object) and must include a `$type` field indicating the variant type. Because the data must be an object, unions can not reference `token` (which would correspond to string data).
+A closed union with only a single type referenced is similar to `ref`. The main difference is that the data is required to include the `$type` field.
+
+An empty closed union (a union with an empty refs list and the `closed` flag set to true) is an invalid schema.
+
+The schema definitions pointed to by a `union` must have type `object` or types with a clear mapping to an object, like a `record`. All the variants must be represented by a CBOR map (or JSON Object) and must include a `$type` field indicating the variant type. Because the data must be an object, unions can not reference `token` (which would correspond to string data).
```
Comment (dholms) [inline src/app/[locale]/specs/lexicon/en.mdx]: I would rejigger both of these paragraphs:

The schema definitions referenced to by a `union` must have type `object` or `record`. All the variants must be represented by a CBOR map (or JSON Object) and must include a `$type` field indicating the variant type.  A `union` can not point to types `query`, `procedure`, `subscription`, `unknown`, `blob`, `cid-link`, `array`, `params`, `ref`, `union`, `token` or any other currently defined types.

dholms (approved):
overall lookin good!

src/app/[locale]/specs/lexicon/en.mdx
```diff
@@ -206,23 +206,33 @@ Type-specific fields:
...
+Refs are a mechanism for re-using a schema definition in multiple places. The intended use is to reduce duplication of definitions for types like `object` or a `string` with specific restrictions.
+
+The `ref` string can be a global reference to a Lexicon type definition (an NSID, optionally with a `#`-delimited name indicating a definition other than `main`), or can indicate a local definition within the same Lexicon file (a `#` followed by a name).
+
+When a `ref` points to an `object` type, the type of the encoded data is unambiguous. This means that the `$type` field should not be included in encoded data as a discriminator. This does not apply to references to `record` types, which always include the `$type` discriminator.
+
+A `ref` can not point to a `token` type.
+
+Refs can not be declared as top-level named types in a schema `defs` array. This means that a `ref` can not point to another `ref`, nor a `union`.
```
Comment (matthieusieben) [inline src/app/[locale]/specs/lexicon/en.mdx]:
```suggestion
Refs can not be declared as top-level named types in a schema `defs` map. This means that a `ref` can not point to another `ref`, nor a `union`. They can reference container types that contain `ref` types.
```

src/app/[locale]/specs/lexicon/en.mdx
```diff
@@ -235,21 +245,31 @@ Unions represent that multiple possible types could be present at this location
...
+A closed union with only a single type referenced is similar to `ref`. The main difference is that the data is required to include the `$type` field.
+
+An empty closed union (a union with an empty refs list and the `closed` flag set to true) is an invalid schema.
+
+The schema definitions pointed to by a `union` must have type `object` or types with a clear mapping to an object, like a `record`. All the variants must be represented by a CBOR map (or JSON Object) and must include a `$type` field indicating the variant type. Because the data must be an object, unions can not reference `token` (which would correspond to string data).
+
+To be more complete, `union` references must point to a type `object` or `record`. They can not point to types `query`, `procedure`, `subscription`, `unknown`, `blob`, `cid-link`, `array`, `params`, `ref`, `union`, or any other currently defined types.
+
+Unions can not be declared as top-level named types in a schema `defs` array. This means that a `union` can not reference another `union`, nor a `ref`.
```
Comment (matthieusieben) [inline src/app/[locale]/specs/lexicon/en.mdx]:
```suggestion
Unions can not be declared as top-level named types in a schema `defs` map. This means that a `union` can not reference another `union`, nor a `ref`.
```

src/app/[locale]/specs/lexicon/en.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/lexicon/en.mdx]: they were mentioned in the paragraph just above, but agree it is best to have them in the same list.

src/app/[locale]/specs/lexicon/en.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/lexicon/en.mdx:36]: yeah, will do a separate PR for `permission` and `permission-set` when we add the specs for those (hopefully very soon!)

PR #465, opened 2025-11-17
specs: permissions

bnewbold:
mostly based on:

- https://github.com/bluesky-social/proposals/tree/main/0011-auth-scopes
- https://github.com/bluesky-social/atproto/discussions/4118

review and clarifications needed on:

- behavior of `rpc` permissions and `aud` in permission sets: do we allow DID refs in `aud` in permission sets at all? what happens if both `aud` is defined and `inheritAud: true`? what if `inheritAud: true` but no `aud` is defined on the `include`?
- IIRC we want to allow `blob` resource in permission sets, but not `account` or `identity`. is that right?
- what is the status of `identity:*` and the "submit" action? we sort of pulled back from that in the August update
- should we just not specify any functionality which isn't implemented yet? eg `account:status`
- should we add 'description' to the permission-set type? as a lexicon-internal field and dev docs, not user-facing. mostly for consistency with all other definition types. I didn't include it in the Go implementation, and went out of the way to mention this in this spec

### DIFF
--- a/src/app/[locale]/specs/lexicon/en.mdx
+++ b/src/app/[locale]/specs/lexicon/en.mdx
@@ -247,9 +247,9 @@ Additional `permission` fields depend on the resource type being declared. The s
     - `aud` (string, semi-required): remote service ("audience"). In the context of permission sets, can not have fixed DID service references. Supports wildcard (`*`), though `aud` and `lxm` can not both be wildcard. Required unless `inheritAud` is true. Effectively must be either wildcard, or undefined with `inheritAud` set to true.
     - `inheritAud` (boolean, optional): if true, `aud` does not need to be defined.
 
-Note that the `blob`, `account`, and `identity` resources types are not supported in Lexicon schema context.
+Note that the `blob`, `account`, and `identity` resources types are not supported in Lexicon schema context, and should be ignored if included.
 
-Permission declarations with unknown resource types or unknown parameters (fields) must be ignored by services implementing access control. This is because the set of permissions is expected to evolve over time, and additional fields might attenuate the permission granted. It is unsafe to ignore unknown fields when parsing permissions for access control.
+Permission declarations with unsupported resource types or parameters (fields) must be ignored by services implementing access control. This is because the set of permissions is expected to evolve over time, and additional fields might attenuate the permission granted. It is unsafe to ignore unknown fields when parsing permissions for access control.
 
 See [Permission specification](/specs/permission) for more details. Note that it is possible to construct a valid lexicon `permission` type which is invalid or has undefined behavior as an auth permission.
 
--- a/src/app/[locale]/specs/permission/en.mdx
+++ b/src/app/[locale]/specs/permission/en.mdx
@@ -99,7 +99,7 @@ rpc:app.example.moderation.createReport?aud=*
   "type": "permission",
   "resource": "rpc",
   "lxm": ["app.example.moderation.createReport"],
-  "aud": ["*"]
+  "aud": "*"
 }
 
 # Call any RPC method on a specific service (not allowed within permission set)
@@ -206,13 +206,12 @@ service:did:web:com.example#type?key=val
 resource:
 resource:?
 resource:&
+resource?
 ```
 
 Examples which do not match the general syntax:
 
 ```
-resource?
-resource&
 resource:positional?key=québec
 emoji:☺️
 ```
@@ -270,7 +269,7 @@ The below is an example permission set Lexicon. The full syntax is described in 
         {
           "type": "permission",
           "resource": "rpc",
-          "aud": ["*"],
+          "aud": "*",
           "lxm": ["app.example.getFeedSkeleton"]
         }
       ]


axfelix (approved):

src/app/[locale]/specs/lexicon/en.mdx
```diff
@@ -60,21 +62,26 @@ The primary types are:
...
 - `subscription`: Event Stream (WebSocket)
 - `record`: describes an object that can be stored in a repository record
+- `permission-set`: describes a bundle of auth permissions
 
-Each primary definition schema object includes these fields:
+Most of the primary definition schema object includes these fields:
 
 - `type` (string, required): the type value (eg, `record` for records)
-- `description` (string, optional): short, usually only a sentence or two
+- `description` (string, optional): short, usually only a sentence or two. `permission-set` does not include this field (see below)
```
Comment (matthieusieben) [inline src/app/[locale]/specs/lexicon/en.mdx]: TS implementation does allow `description` here (but does nothing with it)

src/app/[locale]/specs/lexicon/en.mdx
```diff
@@ -208,6 +231,34 @@ Note that unlike `object`, there is no `nullable` field on `params`.
 
 `params` can not be declared as a top-level named type in a schema `defs` array. This means that they can not be referenced by a `ref` or `union`.
 
+### `permission`
+
+This is a limited-scope type which is only ever used for the `permissions` field of the `permission-set` primary type, and should not appear in other schema contexts. They map to auth permissions.
```
Comment (matthieusieben) [inline src/app/[locale]/specs/lexicon/en.mdx]:
```suggestion
This is a limited-scope type which is only ever used for the `permissions` field of the `permission-set` primary type, and should not appear in other schema contexts. They map to [auth permissions](specs/permission).
```

src/app/[locale]/specs/lexicon/en.mdx
```diff
@@ -208,6 +231,34 @@ Note that unlike `object`, there is no `nullable` field on `params`.
...
+
+Type-specific fields:
+
+- `resource` (string, required): indicates what account resource is being accessed
```
Comment (matthieusieben) [inline src/app/[locale]/specs/lexicon/en.mdx]:
```suggestion
- `resource` (string, required): indicates what account resource is being granted access to
```

src/app/[locale]/specs/lexicon/en.mdx
```diff
@@ -208,6 +231,34 @@ Note that unlike `object`, there is no `nullable` field on `params`.
...
+
+Additional `permission` fields depend on the resource type being declared. The set of resources may expand over time, but currently includes:
+
+- `repo`: public repository write permissions
```
Comment (matthieusieben) [inline src/app/[locale]/specs/lexicon/en.mdx:242]: Just to make things a little bit more explicit:

```suggestion
- public repository write permissions (`"resource": "repo"`)
```

src/app/[locale]/specs/lexicon/en.mdx
```diff
@@ -208,6 +231,34 @@ Note that unlike `object`, there is no `nullable` field on `params`.
...
+    - `collection` (array of strings, required): indicates record types. Must include at least one element (not empty array). Supports wildcard (`*`), but does not support partial-NSID glob patterns (`com.example.*`)
```
Comment (matthieusieben) [inline src/app/[locale]/specs/lexicon/en.mdx]: I would add a note here:

Only collections under the same authority as the permission-set can be granted access to. This means that `*` here is basically the same as `<permission-set-group-nsid>.*`

src/app/[locale]/specs/lexicon/en.mdx
```diff
@@ -208,6 +231,34 @@ Note that unlike `object`, there is no `nullable` field on `params`.
...
+    - `action` (array of strings, optional): limits set of record operations. Values must be one or `create`, `update`, `delete`, case-sensitive.
```
Comment (matthieusieben) [inline src/app/[locale]/specs/lexicon/en.mdx]:
```suggestion
    - `action` (array of unique strings, optional): limits set of record operations. Values must be one or `create`, `update`, `delete`, case-sensitive.
```

src/app/[locale]/specs/lexicon/en.mdx
Comment (matthieusieben) [inline src/app/[locale]/specs/lexicon/en.mdx]:
```suggestion
    - `collection` (array of unique strings, required): indicates record types. Must include at least one element (not empty array). Supports wildcard (`*`), but does not support partial-NSID glob patterns (`com.example.*`)
```

src/app/[locale]/specs/lexicon/en.mdx
```diff
@@ -208,6 +231,34 @@ Note that unlike `object`, there is no `nullable` field on `params`.
...
+- `rpc`: remote API calls via proxying or service token generation
+    - `lxm` (array of strings, required): allowed XRPC endpoints, as NSIDs. Must include at least one element (not empty array). Supports wildcard (`*`), but does not support partial-NSID glob patterns (`com.example.*`)
+    - `aud` (string, semi-required): remote service, identified by DID service reference (DID followed by required service type fragment, eg `did:web:api.example.com#srvtype`). Supports wildcard (`*`). Required unless `inheritAud` is true.
```
Comment (matthieusieben) [inline src/app/[locale]/specs/lexicon/en.mdx]: I would add a note here: cannot be `*` when `lxm` is `*`

src/app/[locale]/specs/lexicon/en.mdx
```diff
@@ -208,6 +231,34 @@ Note that unlike `object`, there is no `nullable` field on `params`.
...
+- `blob`: media file upload permissions
+    - `accept` (array of strings, required): mimetypes or partial mimetype glob patterns (`*/*` or `text/*` for example). Same syntax as the `accept` field in the `blob` lexicon type.
+- `account`: hosted account management
+    - `attr` (string, required): an aspect or component of account. Wildcard is not supported.
+    - `action` (string, optional): degree of control. Currently supports `read` or `manage`
+- `identity`: network identity (DID) management
+    - `attr` (string, required): an aspect or component of account. May be be wildcard (`*`)
+    - `action` (string, optional): degree of control. Currently supports `manage` or `submit`
+
+See Permission specification for more details. Note that it might be possible to construct a valid lexicon `permission` type which is invalid or has undefined behavior as an auth permission.
```
Comment (matthieusieben) [inline src/app/[locale]/specs/lexicon/en.mdx]: We should add a paragraph here stating when a system that interprets these `permission` items does not recognize a particular field, it **MUST** ignore the permission object, as if it was not present. We should also state that AS must not trigger an error to allow clients to adopt newer specs without breaking interaction with AS on older specs (the client can check the received permissions in the token response).

e.G.

```json
{
  "resource": "account",
  "attr": "email",
  "action": ["manage"],
  "foo": "bar"
}
``` 

```json
{
  "resource": "account",
  "attr": "email",
  "action": ["manage", "spam"]
}
``` 

Must all be silently and entirely ignored.

src/app/[locale]/specs/permission/en.mdx
```diff
@@ -0,0 +1,347 @@
+export const metadata = {
+  title: 'Permissions',
+  description:
+    'Auth Permissions for Account Resources.',
+}
+
+# Permissions
+
+Authorization to atproto resources is described and granted using "permissions". A group of a permissions related to a specific Lexicon namespace (record types and API endpoints) can be bundled together as a "permission set". Both are used in the context of [OAuth](/specs/oauth) to grant client software access to account resources on a PDS. For example, the ability to write records of specific types to the user's public repository, or make authenticated API requests to remote services. Developers declare the permissions their app requires to function, and end users are shown the permissions when granting access to the app.
```
Comment (matthieusieben) [inline src/app/[locale]/specs/permission/en.mdx]: > Authorization to atproto resources

This reads weiredly to me, like the phrase lacks a proper object (but it might just be me).

Maybe: "Access to atproto resources"

src/app/[locale]/specs/lexicon/en.mdx
Comment (matthieusieben) [inline src/app/[locale]/specs/lexicon/en.mdx]:
```suggestion
    - `collection` (array of unique strings, required): indicates record types. Must include at least one element (not empty array). Wildcard (`*`) or partial-NSID glob patterns (`com.example.*`) are **NOT** allowed.
```

Wildcards are actually **NOT** allowed here. The reason is that permission-sets can only reference NSIDs **under** (as in, same domain or strict subdomain) the same authority as the permission-set itself.

Any permission item containing `*` here **must** be rejected (TS implementation will ignore those)

We _could_, in the future, relax this to allow `*` here to mean `<nsid-group>.*` but this makes implementation harder since we don't allow partial wildcards in regular permissions.

src/app/[locale]/specs/lexicon/en.mdx
```diff
@@ -208,6 +231,34 @@ Note that unlike `object`, there is no `nullable` field on `params`.
...
+
+- `resource` (string, required): indicates what account resource is being accessed
+
+Additional `permission` fields depend on the resource type being declared. The set of resources may expand over time, but currently includes:
+
+- `repo`: public repository write permissions
+    - `collection` (array of strings, required): indicates record types. Must include at least one element (not empty array). Supports wildcard (`*`), but does not support partial-NSID glob patterns (`com.example.*`)
+    - `action` (array of strings, optional): limits set of record operations. Values must be one or `create`, `update`, `delete`, case-sensitive.
+- `rpc`: remote API calls via proxying or service token generation
+    - `lxm` (array of strings, required): allowed XRPC endpoints, as NSIDs. Must include at least one element (not empty array). Supports wildcard (`*`), but does not support partial-NSID glob patterns (`com.example.*`)
```
Comment (matthieusieben) [inline src/app/[locale]/specs/lexicon/en.mdx]:
```suggestion
    - `lxm` (array of strings, required): allowed XRPC endpoints, as NSIDs. Must include at least one element (not empty array). Wildcard (`*`) or partial-NSID glob patterns (`com.example.*`) are **NOT** allowed.
```


Wildcards are actually NOT allowed here. The reason is that permission-sets can only reference NSIDs under (as in, same domain or strict subdomain) the same authority as the permission-set itself.

src/app/[locale]/specs/lexicon/en.mdx
Comment (matthieusieben) [reply src/app/[locale]/specs/lexicon/en.mdx]: See here for the implemented logic:

https://github.com/bluesky-social/atproto/blob/b7bc95d6abe1136d3ad23b2ff23f6a9991bb373a/packages/oauth/oauth-scopes/src/scopes/include-scope.ts#L65-L118

src/app/[locale]/specs/lexicon/en.mdx
```diff
@@ -208,6 +231,34 @@ Note that unlike `object`, there is no `nullable` field on `params`.
...
+    - `aud` (string, semi-required): remote service, identified by DID service reference (DID followed by required service type fragment, eg `did:web:api.example.com#srvtype`). Supports wildcard (`*`). Required unless `inheritAud` is true.
+    - `inheritAud` (boolean, optional): if true, `aud` does not need to be defined.
+- `blob`: media file upload permissions
+    - `accept` (array of strings, required): mimetypes or partial mimetype glob patterns (`*/*` or `text/*` for example). Same syntax as the `accept` field in the `blob` lexicon type.
+- `account`: hosted account management
+    - `attr` (string, required): an aspect or component of account. Wildcard is not supported.
+    - `action` (string, optional): degree of control. Currently supports `read` or `manage`
+- `identity`: network identity (DID) management
+    - `attr` (string, required): an aspect or component of account. May be be wildcard (`*`)
+    - `action` (string, optional): degree of control. Currently supports `manage` or `submit`
```
Comment (matthieusieben) [inline src/app/[locale]/specs/lexicon/en.mdx]: There are not allowed in permission sets.

```suggestion
`account` and `identity` permissions **cannot** be granted through permission sets and **must** be ignored. This is due to the fact that permission sets are used to conceal multiple app specific permission under a single item in the consent screen. Because of this only resources (record types and lexicon methods) under the authority of the Lexicon document author can be listed in permission sets.
```

src/app/[locale]/specs/lexicon/en.mdx
```diff
@@ -208,6 +231,34 @@ Note that unlike `object`, there is no `nullable` field on `params`.
...
+    - `action` (array of strings, optional): limits set of record operations. Values must be one or `create`, `update`, `delete`, case-sensitive.
+- `rpc`: remote API calls via proxying or service token generation
+    - `lxm` (array of strings, required): allowed XRPC endpoints, as NSIDs. Must include at least one element (not empty array). Supports wildcard (`*`), but does not support partial-NSID glob patterns (`com.example.*`)
+    - `aud` (string, semi-required): remote service, identified by DID service reference (DID followed by required service type fragment, eg `did:web:api.example.com#srvtype`). Supports wildcard (`*`). Required unless `inheritAud` is true.
+    - `inheritAud` (boolean, optional): if true, `aud` does not need to be defined.
+- `blob`: media file upload permissions
+    - `accept` (array of strings, required): mimetypes or partial mimetype glob patterns (`*/*` or `text/*` for example). Same syntax as the `accept` field in the `blob` lexicon type.
```
Comment (matthieusieben) [inline src/app/[locale]/specs/lexicon/en.mdx]: Ideally, the `blob` permission would be limited to adding blobs to record collections listed in the permission set. For example

`include:a include:b`

where `a` has `blob` and `b` doesn't, should result in the app not being able to link `b` collections to blobs.

This is not implemented in TS (`blob` appearing anywhere in the permission scopes will be granted to every collection referenced through the scopes).

src/app/[locale]/specs/permission/en.mdx
```diff
@@ -0,0 +1,347 @@
...
+
+Each resource type has a defined set of parameters that can attenuate the permission. A permission can be represented in a string format (for direct use as an OAuth scope), or as a JSON object (for use in permission sets). Permission sets are published as public lexicon schemas, and are limited to lexicon resources under the same NSID hierarchy.
```
Comment (matthieusieben) [inline src/app/[locale]/specs/permission/en.mdx]:
```suggestion
Each resource type has a defined set of parameters that can attenuate the permission. A permission can be represented in a string format (for direct use as an OAuth scope), or as a JSON object (for use in permission sets). Permission sets are published as public lexicon schemas, and are constrained to describe access to resources under the same NSID hierarchy as the permission set document itself.
```

src/app/[locale]/specs/permission/en.mdx
```diff
@@ -0,0 +1,347 @@
...
+
+## Resource Types
+
+Permissions relate to account resources on PDS instances:
```
Comment (matthieusieben) [inline src/app/[locale]/specs/permission/en.mdx]:
```suggestion
Permissions relate to users owned resources on PDS instances:
```

src/app/[locale]/specs/permission/en.mdx
```diff
@@ -0,0 +1,347 @@
...
+- `rpc`: Service Authentication (API calls to external services)
+- `blob`: uploaded media files
+- `identity`: DID and handle
+- `account`: hosting status, email address
+
+There is an additional "meta" resource, `include`, which is used to reference a permission set (see Permission Sets section below).
+
+This section describes these resources in more depth, including all the parameters that can constrain the permission. For each resource, one parameter may be marked as "positional", which impacts the string representation syntax (described in a later section).
+
+Resources and permissions are described in the abstract, not tied to specific PDS API endpoints. For example, `rpc` permissions describe the ability to make remote service API calls using inter-service authentication. This includes both proxied PDS requests (using HTTP headers) and fetching service auth tokens using `com.atproto.server.getServiceAuth`.
```
Comment (matthieusieben) [inline src/app/[locale]/specs/permission/en.mdx]:
```suggestion

Resources and permissions are described in the abstract, not tied to specific PDS API endpoints. For example, `rpc` permissions describe the ability to make remote service API calls using inter-service authentication. This includes both proxied PDS requests (using the `atproto-proxy` HTTP header) and fetching service auth tokens using `com.atproto.server.getServiceAuth`.
```

src/app/[locale]/specs/permission/en.mdx
```diff
@@ -0,0 +1,347 @@
...
+
+### `repo`
+
+Write access to records in the account's public repository. Can be limited to specific record types (collections) or actions (eg update vs delete).
+
+Parameters:
+
+- `collection` (string array, required, positional): NSID of record types. Wildcard (`*`) is allowed and gives access to all records. Partial wildcards are not supported (eg, can not use `com.example.*`)
```
Comment (matthieusieben) [inline src/app/[locale]/specs/permission/en.mdx]:
```suggestion
- `collection` (string array, required, positional): NSID of record types. Wildcard (`*`) is allowed (when the permission is encoded as an oauth scope parameter i.e. `repo:*`) and gives access to all records. Partial wildcards are not supported (eg, can not use `com.example.*`)
```

src/app/[locale]/specs/permission/en.mdx
```diff
@@ -0,0 +1,347 @@
...
+# same as above, with actions being explicit
+repo:app.example.profile?action=create&action=update&action=delete
+
+{
+  "type": "permission",
+  "resource": "repo",
+  "collection": ["app.example.profile"],
+  "action": ["create", "update", "delete"]
+}
+
```
Comment (matthieusieben) [inline src/app/[locale]/specs/permission/en.mdx:59]: We should add an example with multiple collections here as this will help illustrating that the positional can only be used for array types when a single element is present.

src/app/[locale]/specs/permission/en.mdx
```diff
@@ -0,0 +1,347 @@
...
+Parameters:
+
+- `attr` (string, required): a component of account configuration. Wildcard is not supported.
+- `action` (string, optional): degree of control. Currently supports `read` or `manage`. If not specified, default is `read`.
+
+Attributes:
+
+- `email`: account email address. The `read` action makes the email and verification status visible. The `manage` action includes `read`, and also allows changing the email address.
+- `repo`: ability to update entire public repository using a CAR file. The `manage` action allows importing entire CAR files, for example during account migration. The `read` action does nothing.
+- `status`: account status in network. The `manage` action allows deactivating and re-activating the account. The `read` action does nothing. (NOTE: this attribute is not part of PDS reference implementation as of November 2025)
```
Comment (matthieusieben) [inline src/app/[locale]/specs/permission/en.mdx]: IIRC we mentioned this would be performed through the account management interface and never through 3rd party apps/clients.

src/app/[locale]/specs/permission/en.mdx
```diff
@@ -0,0 +1,347 @@
...
+}
+```
+
+## `blob`
+
+Ability to upload media files (blobs) to PDS.
+
+Parameters:
+
+- `accept` (array of strings, required): MIME types or partial MIME type glob patterns (`*/*` or `text/*` for example). Same syntax as the `accept` field in the `blob` lexicon type.
```
Comment (matthieusieben) [inline src/app/[locale]/specs/permission/en.mdx]: note: Should we add a `collections` field here to restrict which collection can be referencing new blobs ?

Example use case would be "allow app X to upload images for your profile; but not for posts"

src/app/[locale]/specs/permission/en.mdx
```diff
@@ -0,0 +1,347 @@
...
+
+Parameters:
+
+- `attr` (string, required): an aspect or component of account. May be be wildcard (`*`), indicating full control of DID document and handle.
+- `action` (string, optional): degree of control, with context dependent on the attribute. Default is `manage`.
+
+Attributes:
+
+- `handle`: ability to update handle. This includes the registration of the handle in the DID document, as well as any domain names controlled by the PDS. The `manage` action gives full control, including PLC directory submission (if supported)
+- `*` (wildcard): full control over DID document and handle. The `manage` action allows updates of individual fields. The `submit` action includes that functionality, as well as the ability to sign and submit arbitrary PLC directory operations. (NOTE: not included in reference PDS implementation as of November 2025)
```
Comment (matthieusieben) [inline src/app/[locale]/specs/permission/en.mdx]: This is quite hard to comprehend, and the reason we dropped the `submit` action altogether in the implementation. The name submit conflicts with `submitPlcOperation`, which _does not_ actually allow to "sign and submit arbitrary PLC ops".

`com.atproto.identity.submitPlcOperation` contains guard rails that helps cautions users to avoid making mistakes (as long as they only make changes **exclusively** through that endpoint), but those guard rails **do not** protect the "identity" resource itself, as it is, by design, not possible to prevent a client that has "only" the `manage` permission, from injecting its own rotation key and taking over the account out of band.

Because there is no difference in the way these parameters **effectively** protect the user's identity resource, any distinction we'd make here can only be misleading for both client devs and users.

I strongly believe that we should remove this from the spec too.

src/app/[locale]/specs/permission/en.mdx
```diff
@@ -0,0 +1,347 @@
...
+```
+resource?
+resource&
+resource:positional?key=québec
+emoji:☺️
+```
+
+## Permission Sets
+
+Full-featured client apps require a large number of granular permissions to function: dozens or even hundreds of individual permissions. This presents a user experience challenge, as long lists of permissions are difficult to parse, and a developer experience issue, as defining and maintaining these lists is toilsome.
```
Comment (matthieusieben) [inline src/app/[locale]/specs/permission/en.mdx]:
```suggestion
Full-featured client apps require a large number of granular permissions to function: dozens or even hundreds of individual permissions. This presents a user experience challenge, as long lists of permissions are difficult to display in the consent screen, and a developer experience issue, as defining and maintaining these lists is toilsome.
```

src/app/[locale]/specs/permission/en.mdx
```diff
@@ -0,0 +1,347 @@
...
+
+To simplify permission management, Lexicon designers can define "sets" of permissions as part of the schemas they publish. These permission sets are themselves Lexicon schemas and are referred to by NSID, such as `com.example.authBasicFeatures`.
+
+Auth Servers resolve, authenticate, and process permission-sets dynamically. Sets include human-meaningful names and descriptions that are displayed to end users as part of the auth approval flow. Permission sets are published publicly and can be used by any client developer. (Caching and fallback behaviors for dynamic resolution are discussed below.)
```
Comment (matthieusieben) [inline src/app/[locale]/specs/permission/en.mdx]:
```suggestion
Auth Servers resolve, authenticate, and process permission-sets dynamically. Sets include human-meaningful names and descriptions that are displayed to end users as part of the authorization flow. Permission sets are published publicly and can be used by any client developer. (Caching and fallback behaviors for dynamic resolution are discussed below.)
```

src/app/[locale]/specs/permission/en.mdx
```diff
@@ -0,0 +1,347 @@
...
+
+Permission sets are Lexicon schemas and are published and fetched using the Lexicon resolution system, which includes cryptographic authentication. Permission sets can be updated over time as new schemas are added to a namespace (eg, new record types or API endpoints). Auth Servers are expected to maintain a cache of resolved sets, but to re-resolve them periodically. The permissions associated with an Access Token should remain fixed, but when a client refreshes their tokens (obtaining a new access token), the computed permissions for the session may be updated to reflect changes to sets requested by the client.
+
+This adds an intentional degree of temporal dynamism to auth sessions involving Permission sets. Lexicon designers can define new resources (eg, record types), update published sets to include permissions to those resources. Client software can then be updated to take advantage of those new lexicons, without requiring users to re-authenticate their sessions.
```
Comment (matthieusieben) [inline src/app/[locale]/specs/permission/en.mdx]: We might want to add, either here or in the oauth spec, that the oauth session **must** be bound to the **exact** scopes requested during the initial request and that this is the only way to add some dynamism.

src/app/[locale]/specs/permission/en.mdx
```diff
@@ -0,0 +1,347 @@
...
+### Namespace Authority
+
+Permission sets are limited to expressing permissions that reference resources under the same NSID namespace as the set itself.
+
+Authority is based on the relative structure of NSIDs, without "siblings" or special namespaces. Broadly, this ensures that sets can not request permissions across namespaces. Specifically:
+
+- Permission sets can address resources in the same NSID "group"; or "children" (sub-domains), recursively deep
+- can not address "sibling groups" or "parents" in the NSID hierarchy
+
+For example, the set `app.example.feed.authOnlyPost` could include permissions to `app.example.feed.post` records and making `app.example.feed.getPostThread` API endpoint requests to remote services. But it could not grant permissions to `app.example.actor.profile`. A permission set `app.example.authFull`, which is a level up in the hierarchy, could include permissions to all these resources, or even further down the hierarchy.
```
Comment (matthieusieben) [inline src/app/[locale]/specs/permission/en.mdx:305]: We should add

> AS and RS must silently ignore (i.e. act as if they were not present) any `permission` item that contains unknown or invalid parameter keys or values.

src/app/[locale]/specs/permission/en.mdx
```diff
@@ -0,0 +1,347 @@
...
+
+### Resolution and Caching
+
+Permission set Lexicons need to be resolved by auth servers (eg, PDS instances). The existing [Lexicon publication and resolution system](/specs/lexicon) describes how schemas are published and verified.
+
+To reduce network traffic and increase resiliency to outages, auth servers are expected to cache of permission set resolution. Caches may be shared across all accounts and sessions on an Auth Server.
+
+Permission set schemas should be cached with long "expiration" times but shorter "stale" times. Stale Lexicons should get updated (aka, attempt resolution refresh), but if resolution fails, the previously existing stale values can be used. The recommended "stale" lifetime is 24 hours, and this is intended as a firm upper bound on cache lifetime. The firm lower bound on cache lifetime is that of access token lifetimes, meaning 15-30 minutes. The recommended "expiration" lifetime is 90 days, but this is not a firm bound.
```
Comment (matthieusieben) [inline src/app/[locale]/specs/permission/en.mdx]: > The recommended "expiration" lifetime is 90 days, but this is not a firm bound.

This is touchy as it means that apps could start braking after a while.

I think that we should mention that permissions sets cannot be marked as "expired" for as long as there are active oauth session relying on them. It's fine to prevent new session from being created if they reference expired permission sets but existing sessions should not break.

src/app/[locale]/specs/permission/en.mdx
```diff
@@ -0,0 +1,347 @@
...
+
+At the start of a session, if a permission set can not be resolved (and is not already in a local cache), the auth request will fail.
+
+As an optional mechanism to trigger a cache purge, client instances can set an updated `software_version` field in their client metadata document. If the auth server sees a client version change, it should ensure that any permission sets referenced by the client would be refreshed. The motivation is to ensure that roll-out of new client software that requires updated sets can rely on auth servers giving the updated permissions within a timeframe shorter than the 24 hour "stale" cache lifetime.
```
Comment (matthieusieben) [inline src/app/[locale]/specs/permission/en.mdx]: We don't do this, and never really reach any consensus on this. Can we leave it out from the spec until we do reach a consenus ?

src/app/[locale]/specs/permission/en.mdx
```diff
@@ -0,0 +1,347 @@
...
+
+
+## Usage and Implementation Guidelines
+
+The currently defined resources and parameters do not make use of all of the syntax flexibility. Implementations might decide to focus on currently used syntax instead of the general syntax.
+
+Auth Servers may chose to use a "Lexicon Aggregator" service to offload lexicon resolution of permission sets. Such services would monitor a full-network firehose to detect updated permission sets. They could contain security logic to protect against NSID domain hijacking or adversarial changes to permission set scope.
```
Comment (matthieusieben) [inline src/app/[locale]/specs/permission/en.mdx]: We could be more specific and state that authorization servers are allowed to map an `include:<nsid>` to a different set of permissions than those included in the permission-set.

Clients can inspect the `scope` parameter of the token response to verify that they got the right permissions (OAuth spec says that the AS *must* return the `scope` parameter if it does not match the requested one)

src/app/[locale]/specs/permission/en.mdx
Comment (dholms) [reply src/app/[locale]/specs/permission/en.mdx]: or "Authorization for modifying atproto resources"

src/app/[locale]/specs/permission/en.mdx
Comment (dholms) [reply src/app/[locale]/specs/permission/en.mdx]: We should what's meant by the parenthetical more explicitly^^

src/app/[locale]/specs/permission/en.mdx
```diff
@@ -0,0 +1,347 @@
...
+repo:*
+
+{
+  "type": "permission",
+  "resource": "repo",
+  "collection": ["*"]
+}
+
+# delete only permission on all record types
+repo:*?action=delete
```
Comment (dholms) [inline src/app/[locale]/specs/permission/en.mdx:73]: should we add the JSON representation of this so that it matches the other examples?

src/app/[locale]/specs/permission/en.mdx
```diff
@@ -0,0 +1,347 @@
...
+```
+
+### `rpc`
+
+The ability to make authenticated API requests to remote services. This includes both requesting service auth tokens (JWTs) using the `com.atproto.server.getServiceAuth` endpoint on the PDS, and requests proxied via the PDS.
+
+The permission is parameterized by the remote endpoint (`lxm`, short for "Lexicon Method") and the identity of the remote service (the audience, `aud`). Permissions must be restricted by one or the other: the ability to call any API endpoint on any service can not be declared with a single permission.
```
Comment (dholms) [inline src/app/[locale]/specs/permission/en.mdx]:
```suggestion
The permission is parameterized by the remote endpoint (`lxm`, short for "Lexicon Method") and the identity of the remote service (the audience, `aud`). Permissions must be restricted by at least one of these parameters. In other words, the ability to call any API endpoint on any service can not be declared with a single permission.
```

src/app/[locale]/specs/permission/en.mdx
```diff
@@ -0,0 +1,347 @@
...
+
+When included in a permission set, the audience parameter may be inherited from the overall set.
+
+Parameters:
+
+- `lxm` (string array, required, positional): NSID of API endpoints. Wildcard (`*`) is allowed and gives access to all endpoints. Partial wildcards are not supported (eg, can not use `com.example.*`).
+- `aud` (string, semi-required): audience of API requests, as a DID service reference: DID followed by required service type fragment, eg `did:web:api.example.com#srvtype`). Supports wildcard (`*`). Required in granular string representation, and required in lexicon JSON unless `inheritAud` is true.
+- `inheritAud` (boolean, optional): if true, `aud` does not need to be defined.
```
Comment (dholms) [inline src/app/[locale]/specs/permission/en.mdx]: maybe move the note from about down to here & note that this is only relevant in the context of a permission set

src/app/[locale]/specs/permission/en.mdx
Comment (dholms) [reply src/app/[locale]/specs/permission/en.mdx]: this isn't implemented anywhere right? I think we hold off

src/app/[locale]/specs/permission/en.mdx
Comment (dholms) [reply src/app/[locale]/specs/permission/en.mdx]: yeah i agree with this^^

src/app/[locale]/specs/permission/en.mdx
Comment (dholms) [inline src/app/[locale]/specs/permission/en.mdx]:
```suggestion
- `accept` (array of strings, required, positional): MIME types or partial MIME type glob patterns (`*/*` or `text/*` for example). Same syntax as the `accept` field in the `blob` lexicon type.
```

src/app/[locale]/specs/permission/en.mdx
```diff
@@ -0,0 +1,347 @@
...
+
+## `account`
+
+Control of PDS account hosting details, such as (private) account email.
+
+Permissions of this type can not be included in permission sets, and must be requested directly by client apps.
+
+Parameters:
+
+- `attr` (string, required): a component of account configuration. Wildcard is not supported.
```
Comment (dholms) [inline src/app/[locale]/specs/permission/en.mdx]:
```suggestion
- `attr` (string, required, positional): a component of account configuration. Wildcard is not supported.
```

src/app/[locale]/specs/permission/en.mdx
```diff
@@ -0,0 +1,347 @@
...
+
+## `identity`
+
+Control over network identity, meaning the account DID document and handle. Note that the PDS might not be able to facilitate identity changes if it does not have control over the DID document (for example, when the account uses `did:web`).
+
+Permissions of this type can not be included in permission sets, and must be requested directly by client apps.
+
+Parameters:
+
+- `attr` (string, required): an aspect or component of account. May be be wildcard (`*`), indicating full control of DID document and handle.
```
Comment (dholms) [inline src/app/[locale]/specs/permission/en.mdx]:
```suggestion
- `attr` (string, required, positional): an aspect or component of account. May be be wildcard (`*`), indicating full control of DID document and handle.
```

src/app/[locale]/specs/permission/en.mdx
Comment (dholms) [reply src/app/[locale]/specs/permission/en.mdx]: yeah I think it's worth noting the security implications of avoiding bloated consent screens

src/app/[locale]/specs/permission/en.mdx
```diff
@@ -0,0 +1,347 @@
...
+- Permission sets can address resources in the same NSID "group"; or "children" (sub-domains), recursively deep
+- can not address "sibling groups" or "parents" in the NSID hierarchy
+
+For example, the set `app.example.feed.authOnlyPost` could include permissions to `app.example.feed.post` records and making `app.example.feed.getPostThread` API endpoint requests to remote services. But it could not grant permissions to `app.example.actor.profile`. A permission set `app.example.authFull`, which is a level up in the hierarchy, could include permissions to all these resources, or even further down the hierarchy.
+
+### Resolution and Caching
+
+Permission set Lexicons need to be resolved by auth servers (eg, PDS instances). The existing [Lexicon publication and resolution system](/specs/lexicon) describes how schemas are published and verified.
+
+To reduce network traffic and increase resiliency to outages, auth servers are expected to cache of permission set resolution. Caches may be shared across all accounts and sessions on an Auth Server.
```
Comment (dholms) [inline src/app/[locale]/specs/permission/en.mdx]:
```suggestion
To reduce network traffic and increase resiliency to outages, auth servers are expected to cache permission set resolution. Caches may be shared across all accounts and sessions on an Auth Server.
```

src/app/[locale]/specs/lexicon/en.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/lexicon/en.mdx:242]: hrm, I feel like this JSON-style syntax is less readable for the whole list.

One thing this makes me feel is that we should maybe have minimal examples of each of the definition types in this overall lexicon spec file.

src/app/[locale]/specs/lexicon/en.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/lexicon/en.mdx]: I added a section about this here in the lexicon document, and in the permission spec.

src/app/[locale]/specs/permission/en.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/permission/en.mdx]: I went with "Control over atproto resources"

src/app/[locale]/specs/permission/en.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/permission/en.mdx]: I will remove for now because it isn't supported.

But we should revisit in the future. Eg, will we drop support for that functionality from bsky app when we switch to OAuth? All apps should be on a level playing field.

src/app/[locale]/specs/permission/en.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/permission/en.mdx]: This was me not remembering what we had settled on.

I pulled out all mention of `action`.

src/app/[locale]/specs/permission/en.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/permission/en.mdx:73]: We don't allow wildcard in the context of permission sets, so I think providing JSON representation would be confusing here. I added a clarifying comment and removed JSON from another disallowed example above.

src/app/[locale]/specs/permission/en.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/permission/en.mdx]: I re-worded a bit, but kept a reference/mention on the `aud` line

src/app/[locale]/specs/permission/en.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/permission/en.mdx:305]: I added a paragraph about this, but in the section just above this

src/app/[locale]/specs/permission/en.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/permission/en.mdx]: Removed the paragraph about the `software_version` mechanism, and added a paragraph mentioning it under "possible future changes".

src/app/[locale]/specs/lexicon/en.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/lexicon/en.mdx]: I think this might be nice to add in the future, but we didn't include in this design, and I don't think we should iterate right now.

I agree there is some permission-leakage possible across permission sets, but I don't feel a large urgency around this right now.

src/app/[locale]/specs/permission/en.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/permission/en.mdx]: I agree we shouldn't add functionality right now, let's just keep blobs broad.

src/app/[locale]/specs/permission/en.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/permission/en.mdx]: I think the second part of that is an OAuth semantic and should live in that spec (if it isn't already).

I'm not sure how to word the other part, or if we even really need to include it for now. In some sense the Auth Server can always do whatever it wants to protect the security of users.

I think I would lean towards not expanding on this in a very general way until we have something more concrete to describe or recommend.

bnewbold:
- pushed some edits based on review feedback which I forgot to commit last night
- resolved the two open questions above (no blobs in permission-sets; no DID service refs in permission sets)
- resolved merge conflicts with `main`

src/app/[locale]/specs/lexicon/en.mdx
```diff
@@ -208,6 +229,30 @@ Note that unlike `object`, there is no `nullable` field on `params`.
...
+- `rpc`: remote API calls via proxying or service token generation
+    - `lxm` (array of strings, required): allowed XRPC endpoints, as NSIDs. Must include at least one element (not empty array). Wildcard (`*`) is not supported in Lexicon schema context.
+    - `aud` (string, semi-required): remote service ("audience"). In the context of permission sets, can not have fixed DID service references. Supports wildcard (`*`), though `aud` and `lxm` can not both be wildcard. Required unless `inheritAud` is true. Effectively must be either wildcard, or undefined with `inheritAud` set to true.
+    - `inheritAud` (boolean, optional): if true, `aud` does not need to be defined.
+
+Note that the `blob`, `account`, and `identity` resources types are not supported in Lexicon schema context.
+
+Permission declarations with unknown resource types or unknown parameters (fields) must be ignored by services implementing access control. This is because the set of permissions is expected to evolve over time, and additional fields might attenuate the permission granted. It is unsafe to ignore unknown fields when parsing permissions for access control.
```
Comment (matthieusieben) [inline src/app/[locale]/specs/lexicon/en.mdx]: We should make it very explicit that `blob`, `account`, and `identity` must be ignored if present. Either like suggested here, or as an additional sentense in the previous paragraph.

```suggestion
Permission declarations with unknown or disallowed resource types or unknown parameters (fields) must be ignored by services implementing access control. This is because the set of permissions is expected to evolve over time, and additional fields might attenuate the permission granted. It is unsafe to ignore unknown fields when parsing permissions for access control.
```

src/app/[locale]/specs/permission/en.mdx
```diff
@@ -0,0 +1,331 @@
...
+Examples:
+
+```
+# Upload any type of blob
+blob:*/*
+
+# Upload video or html
+blob?accept=video/*&accept=text/html
+```
+
```
Comment (matthieusieben) [inline src/app/[locale]/specs/permission/en.mdx:128]: Should we explicit in these sections that these permissions cannot appear in permission sets (and should be ignored if they do) ?

Note that we might (at some point and based on usage feedback) be tempted to create `com.atproto.` permission sets that *do* contain `account`/`identity` permissions.

src/app/[locale]/specs/permission/en.mdx
```diff
@@ -0,0 +1,331 @@
...
+resource:
+resource:?
+resource:&
+```
+
+Examples which do not match the general syntax:
+
+```
+resource?
+resource&
```
Comment (matthieusieben) [inline src/app/[locale]/specs/permission/en.mdx]: these 2 are actually syntactically fine

src/app/[locale]/specs/permission/en.mdx
```diff
@@ -0,0 +1,331 @@
...
+            "app.example.getPreferences",
+            "app.example.putPreferences",
+            "app.example.getAuthorFeed",
+            ...
+          ]
+        },
+        {
+          "type": "permission",
+          "resource": "rpc",
+          "aud": ["*"],
```
Comment (matthieusieben) [inline src/app/[locale]/specs/permission/en.mdx]: `aud` cannot be an array:

https://github.com/bluesky-social/atproto/blob/1ca93f45293ba86281dc542e0dc1e065605fc517/packages/oauth/oauth-scopes/src/scopes/rpc-permission.ts#L55

src/app/[locale]/specs/permission/en.mdx
Comment (matthieusieben) [reply src/app/[locale]/specs/permission/en.mdx]: They parse as:

```ts
// resource?
const first = {
  resource: "resource",
  positional: undefined,
  params: {}
}

// resource&
const second = {
  resource: "resource&",
  positional: undefined,
  params: undefined
}
```

src/app/[locale]/specs/permission/en.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/permission/en.mdx]: whoops, good catch!

src/app/[locale]/specs/permission/en.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/permission/en.mdx:128]: I didn't include those line-level notes for the resource types which aren't supported in permission sets at all, only for the cases where special usage (eg, lxm=*) is specifically disallowed. there is a pretty clear line just above.

PR #468, opened 2025-11-18
start to improve oauth recs

axfelix:
Got a starting point up that incorporates some outstanding feedback/complaints from users. Still need to pull in more content from https://github.com/bluesky-social/atproto/discussions/4118 and weave it into both the guide and the specs page.


src/app/[locale]/guides/oauth/en.mdx
```diff
@@ -76,3 +76,16 @@ There are several design strategies that can be used to "upgrade" a public clien
...
 
 Of these three, the BFF pattern is the currently recommended approach for building confidential-client applications. This is demonstrated in the [cookbook/go-oauth-web-app](https://github.com/bluesky-social/cookbook/tree/main/go-oauth-web-app#readme) and [cookbook/python-oauth-web-app](https://github.com/bluesky-social/cookbook/tree/main/python-oauth-web-app#readme) examples.
+
+## Scopes
+
+The OAuth flow requires permission "scopes" which describe the level of access granted with the session. For now, granular permissions have been implemented on the `bsky.social` hosting service, and are rolling out to the self-hosted PDS distribution. Work on permission sets (including lexicon resolution) is in progress. Users may start to implement more granular permissions -- for example apps only asking for permissions to specific record types or API endpoints. We expect most apps to use the higher-level "permission sets" when they become available, which will provide more accessible language about the permissions being granted.
```
Comment (jimray) [inline src/app/[locale]/guides/oauth/en.mdx]: Maybe link "self-hosted PDS distribution" here: /guides/self-hosting

Comment (jimray) [inline src/app/[locale]/guides/oauth/en.mdx]: Suggest changing "Users may start to implement..." to "Developers"

Comment (jimray) [inline src/app/[locale]/guides/oauth/en.mdx]: I wonder if it's worth adding a sentence or two going into a bit more detail about what OAuth scopes entail. Anyone who's implemented OAuth before will understand, but if this is your first introduction it might be helpful to expand that first sentence; it also helps to disambiguate OAuth from something like app passwords, which just give full access to the third party app. Maybe something like:

"One way OAuth is different from just logging into an app with your username and password is that apps using OAuth are able to request specific permissions. Unless you are building a fully-featured AT Protocol client, you likely don't need full permission every feature of a user's PDS. By only requesting the permissions your app needs, it keeps both your app and user data more secure."

src/app/[locale]/specs/oauth/en.mdx
```diff
@@ -44,7 +47,15 @@ OAuth client software is identified by a globally unique `client_id`. Distinct v
 
 - it must be a well-formed URL, following the W3C URL specification
 - the schema must be `https://`, and there must not be a port number included. Note that there is a special exception for `http://localhost` `client_id` values for development, see details below
-- the path does not need to include `client-metadata.json`, but it is helpful convention
+- the path does not need to include `client-metadata.json`, but it is helpful convention.
```
Comment (jimray) [inline src/app/[locale]/specs/oauth/en.mdx]: Feels weird to recommend this and then recommend `oauth-client-metadata.json` below?

jimray (commented):
Added a few comments, looking good!

src/app/[locale]/specs/oauth/en.mdx
Comment (axfelix) [reply src/app/[locale]/specs/oauth/en.mdx]: Yup, completely agree, was just trying to get that part of my notes into a PR draft I could push up before lunch :)

src/app/[locale]/specs/oauth/en.mdx
```diff
@@ -222,7 +307,7 @@ For web clients, the `redirect_uri` is a HTTPS URL which will be redirected in t
...
+For native clients, the `redirect_uri` may use a custom URI scheme to have the operating system redirect the user back to the app, instead of a web browser. Native clients are also allowed to use an HTTPS URL. Any custom scheme must match the `client_id` hostname in reverse-domain order. The URI scheme must be followed by a single colon (`:`) then a single forward slash (`/`) and then a URI path component. For example, an app with `client_id`  [`https://app.example.com/oauth-client-metadata.json`](https://app.example.com/oauth-client-metadata.json) could have a `redirect_uri` of `com.example.app:/callback`.
```
Comment (bnewbold) [inline src/app/[locale]/specs/oauth/en.mdx:221]: :+1:

src/app/[locale]/specs/oauth/en.mdx
```diff
@@ -189,8 +204,78 @@ The intention is to support flexible scopes based on Lexicon namespaces (NSIDs)
 - `transition:email`: access to the account email address
     - email address (and confirmation status) gets included in response to `com.atproto.server.getSession` endpoint
 
+Granular permissions have so far only been implemented on the `bsky.social` hosting service. The ergonomics of OAuth scopes in AT Protocol are still evolving. The current implementation is documented in the following section.
```
Comment (bnewbold) [inline src/app/[locale]/specs/oauth/en.mdx]: We should not put this note in the specs (fine in a guide)

src/app/[locale]/guides/oauth/en.mdx
```diff
@@ -76,3 +78,16 @@ There are several design strategies that can be used to "upgrade" a public clien
...
+
+## Scopes
+
+The OAuth flow requires permission "scopes" which describe the level of access granted with the session. For now, granular permissions have been implemented on the `bsky.social` hosting service, and are rolling out to the [self-hosted PDS distribution](/guides/self-hosting). Work on permission sets (including lexicon resolution) is in progress. Developers may start to implement more granular permissions -- for example, apps only asking for permissions to specific record types or API endpoints. We expect most apps to use the higher-level "permission sets" when they become available, which will provide more accessible language about the permissions being granted.
+
+The minimal set of OAuth permissions that a third-party AT application should request to sign in with Bluesky are:
+- `atproto`
+- `rpc:app.bsky.actor.getProfile?aud=did:web:api.bsky.app#bsky_appview`
```
Comment (matthieusieben) [inline src/app/[locale]/guides/oauth/en.mdx:93]: this is not strictly necessary as the application (client) could fetch the public profile using credentials of their own (er even unauthenticated using `https://public.api.bsky.app`)

src/app/[locale]/guides/oauth/en.mdx
```diff
@@ -76,3 +78,16 @@ There are several design strategies that can be used to "upgrade" a public clien
...
+
+If you need to interact with any images or video, you will also need:
+- `blob:*/*`
```
Comment (matthieusieben) [inline src/app/[locale]/guides/oauth/en.mdx:96]: This is useless on its own. Being able to upload blobs is only useful if the app will be creating records containing references to those blobs. I would recommend to use the `repo:app.bsky.feed.like` & co, as base, and `blob:*/*` if the app will be linking files to the records they create/update (posts,profile,whatever).

src/app/[locale]/specs/oauth/en.mdx
```diff
@@ -189,8 +204,78 @@ The intention is to support flexible scopes based on Lexicon namespaces (NSIDs)
...
+
+### Scope String Details
+
+Permission scope strings are case-sensitive and use a subset of printable non-whitespace ASCII characters. They include a resource name, followed by an optional "positional" part separated by a colon (:), followed by optional parameters starting with a question mark (?): for example, `rpc:app.bsky.actor.getProfile?aud=did:web:api.bsky.app#bsky_appview` is used to return a DID, and is required along with the `atproto` scope for a minimal login flow that supports signing in with Bluesky but grants no additional permissions.
```
Comment (matthieusieben) [inline src/app/[locale]/specs/oauth/en.mdx]:
```suggestion
Permission scope strings are case-sensitive and use a subset of printable non-whitespace ASCII characters. They include a resource name, followed by an optional "positional" part separated by a colon (:), followed by optional parameters starting with a question mark (?): for example, `rpc:app.bsky.actor.getProfile?aud=did:web:api.bsky.app#bsky_appview` is used to fetch the user's own profile. The `atproto` scope is required for a minimal login flow that supports signing in with Bluesky but grants no additional permissions (authentication only, no authorization).
```

src/app/[locale]/specs/oauth/en.mdx
```diff
@@ -189,8 +204,78 @@ The intention is to support flexible scopes based on Lexicon namespaces (NSIDs)
...
+
+The positional part can contain any characters except for a question mark, including additional colon characters, hash (#), period, etc. and support being percent encoded. Parameters are separated by ampersands (&), have name and value parts separated by equals (=), and support percent-encoding of reserved characters. Array values can be represented by repeated parameters with the same name part.
```
Comment (matthieusieben) [inline src/app/[locale]/specs/oauth/en.mdx]: The following could be misleading:

> any characters except for a question mark

The [OAuth2.0 RFC](https://www.rfc-editor.org/rfc/rfc6749#appendix-A.4) allows any printable ASCII character (no emojis!) excluding space and double quote and backslash.

We should add a phrase that explains that the string after the question mark MUST be a valid URL encoded query string and that some implementation MAY (don't have to) support non-encoded characters.

src/app/[locale]/specs/oauth/en.mdx
```diff
@@ -189,8 +204,78 @@ The intention is to support flexible scopes based on Lexicon namespaces (NSIDs)
...
+
+```
+# Upload any type of blob
+blob:*/*
+
+# Upload video or html
+blob?accept=video/*&accept=text/html
+```
+
+**Account Permissions**: the `attr` parameter is required, and may be positional. The `action` parameter is optional, can be `read` or `manage` (which includes read/write permissions), and defaults to `read`. The initial supported account attributes are `email` (account email address) and `repo` (for importing entire data repositories, as part of account migration). `status` (account hosting status) can be declared, but is not currently implemented anywhere.
```
Comment (matthieusieben) [inline src/app/[locale]/specs/oauth/en.mdx]:
```suggestion
**Account Permissions**: the `attr` parameter is required, and may be positional. The `action` parameter is optional, can be `read` or `manage` (which includes read/write permissions), and defaults to `read`. The initial supported account attributes are `email` (account email address) and `repo` (for importing entire data repositories, as part of account migration).
```

We still need to agree on this with @bnewbold, but we mentioned removing `status` from the spec alltogether.

src/app/[locale]/specs/oauth/en.mdx
```diff
@@ -189,8 +204,78 @@ The intention is to support flexible scopes based on Lexicon namespaces (NSIDs)
...
+# Manage account handle
+identity:handle
+
+# Full control over DID document, eg for account migration
+identity:*
+```
+
 <Note>
-As of August 2025, a more granular permission system is available for use, announced and described [here](https://github.com/bluesky-social/atproto/discussions/4118). The design is considered stable, although it has not yet been formally documented in this specification.
+Note: An earlier version of these examples showed scope `identity:*?action=submit` (allowing submission of pre-signed updates). This scope does not exist in the current implementation, but some equivalent and/or more granular `identity:` scopes may be introduced in the future.
```
Comment (matthieusieben) [inline src/app/[locale]/specs/oauth/en.mdx]: This is great!

src/app/[locale]/guides/oauth/en.mdx
Comment (axfelix) [reply src/app/[locale]/guides/oauth/en.mdx:93]: OK, interesting, I was going off of this anecdata (https://bsky.app/profile/wisp.place/post/3m5tvmsvrr22o) and also wanting to offer an example of a very specific scope. You think this is not an actual edge case?

src/app/[locale]/specs/oauth/en.mdx
Comment (axfelix) [reply src/app/[locale]/specs/oauth/en.mdx]: OK! Sounds like we have some very small blockers but can hopefully get this merged by EOW.

Comment (axfelix) [reply src/app/[locale]/specs/oauth/en.mdx]: Actually, I'll just remove this from the PR for now if it's unused!

src/app/[locale]/guides/oauth/en.mdx
Comment (bnewbold) [reply src/app/[locale]/guides/oauth/en.mdx:93]: You definitely shouldn't need anything other than `atproto` for basic "authentication" (just check that the user is who they say they are, without authorization to any specific resources).

I can't really tell what the wisp dev is saying about getProfile. Other apps should not need getProfile to function *in general*, but wisp specifically might make use of getProfile as part of their overall app experience.

src/app/[locale]/guides/oauth/en.mdx
```diff
@@ -76,3 +78,16 @@ There are several design strategies that can be used to "upgrade" a public clien
...
 
 Of these three, the BFF pattern is the currently recommended approach for building confidential-client applications. This is demonstrated in the [cookbook/go-oauth-web-app](https://github.com/bluesky-social/cookbook/tree/main/go-oauth-web-app#readme) and [cookbook/python-oauth-web-app](https://github.com/bluesky-social/cookbook/tree/main/python-oauth-web-app#readme) examples.
+
+## Scopes
+
+The OAuth flow requires permission "scopes" which describe the level of access granted with the session. For now, granular permissions have been implemented on the `bsky.social` hosting service, and are rolling out to the [self-hosted PDS distribution](/guides/self-hosting). Work on permission sets (including lexicon resolution) is in progress. Developers may start to implement more granular permissions -- for example, apps only asking for permissions to specific record types or API endpoints. We expect most apps to use the higher-level "permission sets" when they become available, which will provide more accessible language about the permissions being granted.
```
Comment (bnewbold) [inline src/app/[locale]/guides/oauth/en.mdx]: I think permissions are fully rolled out to everybody running the reference PDS at this point. I wouldn't mention `bsky.social` here in a way that makes it sounds special or different from any other PDS instance.

bnewbold (approved):
looks good now!

src/app/[locale]/guides/oauth/en.mdx
```diff
@@ -76,3 +81,38 @@ There are several design strategies that can be used to "upgrade" a public clien
...
 
 Of these three, the BFF pattern is the currently recommended approach for building confidential-client applications. This is demonstrated in the [cookbook/go-oauth-web-app](https://github.com/bluesky-social/cookbook/tree/main/go-oauth-web-app#readme) and [cookbook/python-oauth-web-app](https://github.com/bluesky-social/cookbook/tree/main/python-oauth-web-app#readme) examples.
+
+## Scopes
+
+The OAuth flow requires permission "scopes" which describe the level of access granted with the session. For now, granular permissions have been implemented on the `bsky.social` hosting service, and the [self-hosted PDS distribution](/guides/self-hosting). Work on permission sets (including lexicon resolution) is in progress. Developers may start to implement more granular permissions -- for example, apps only asking for permissions to specific record types or API endpoints. We expect most apps to use the higher-level "permission sets" when they become available, which will provide more accessible language about the permissions being granted.
+
+The minimal set of OAuth permissions that a third-party AT application should request to sign in with Bluesky are:
```
Comment (ThisIsMissEm) [inline src/app/[locale]/guides/oauth/en.mdx:89]: It'd be worth saying what this gives you as a client developer

PR #477, opened 2025-12-04
clarify support for XRPC redirects

axfelix:


src/app/[locale]/specs/xrpc/en.mdx
```diff
@@ -187,6 +187,12 @@ Only HTTPS should be used over the open internet.
 
 Care should be taken with personally identifiable information in blobs, such as EXIF metadata. It is currently the *client's* responsibility to strip any sensitive EXIF metadata from blobs before uploading. It would be reasonable for a PDS to help prevent accidental metadata leakage as well; see future changes section below.
 
+## Redirects
+
+XRPC is built on top of HTTP. As such, everything that is not explicitly allowed or forbidden by the XRPC spec falls under the authority of HTTP.
+
+This means that all *un-authenticated* "Query" (HTTP GET) XRPC endpoints, such as `getBlob`, support redirects.
```
Comment (dholms) [inline src/app/[locale]/specs/xrpc/en.mdx]: Unauthenticated "Query" (HTTP GET) XRPC endpoints, such as `getBlob`, support redirects.

An explicit decision has not been made around XRPC support for redirects for "Procedure" (HTTP POST) endpoints and therefore should not be relied on.

HTTP clients should take care to strip `Authorization` headers for cross-origin redirects. Most browsers and stdlib HTTP clients handle this behavior

Comment (bnewbold) [reply src/app/[locale]/specs/xrpc/en.mdx]: I'm plus one to danny's text.

I'd also remove the "everything that is not explicitly allowed or forbidden by the XRPC spec falls under the authority of HTTP". It is philosophically true, and maybe helpful around things like transport encoding, but is framed in a semi-formal way that I think is confusing in specification text.

src/app/[locale]/specs/xrpc/en.mdx
```diff
@@ -187,6 +187,14 @@ Only HTTPS should be used over the open internet.
 
 Care should be taken with personally identifiable information in blobs, such as EXIF metadata. It is currently the *client's* responsibility to strip any sensitive EXIF metadata from blobs before uploading. It would be reasonable for a PDS to help prevent accidental metadata leakage as well; see future changes section below.
 
+## Redirects
+
+XRPC is built on top of HTTP. As such, everything that is not explicitly allowed or forbidden by the XRPC spec falls under the authority of HTTP.
+
+Unauthenticated "Query" (HTTP GET) XRPC endpoints, such as `getBlob`, support redirects. An explicit decision has not been made around XRPC support for redirects for "Procedure" (HTTP POST) endpoints and therefore should not be relied on.
```
Comment (ThisIsMissEm) [inline src/app/[locale]/specs/xrpc/en.mdx:194]: There is now a QUERY verb in HTTP, which is GET with a body. Might be worth making explicit that you're referring to an XPRC query, i.e., a HTTP GET request.

src/app/[locale]/specs/xrpc/en.mdx
```diff
@@ -187,6 +187,14 @@ Only HTTPS should be used over the open internet.
...
+
+HTTP clients should take care to strip Authorization headers for cross-origin redirects. Most browsers and HTTP clients handle this behavior.
```
Comment (ThisIsMissEm) [inline src/app/[locale]/specs/xrpc/en.mdx:196]: Depends on if the request contains include credentials iirc. It's been a while since I've read the `fetch()` algorithm though

src/app/[locale]/specs/xrpc/en.mdx
Comment (axfelix) [reply src/app/[locale]/specs/xrpc/en.mdx]: Hm -- I don't think adding "generally" here would be clarifying in that case, and I agree that the framing here is a bit more sensitive in a spec than it would be in a guide, but I also think this is a very good thing to reinforce in cases like this. Is there a rewrite that still expresses this that you think would fit here?

src/app/[locale]/specs/xrpc/en.mdx
Comment (ThisIsMissEm) [inline src/app/[locale]/specs/xrpc/en.mdx:194]:
```suggestion
Unauthenticated XRPC Query endpoints (sent via HTTP GET requests), such as `getBlob`, do support redirects.

An explicit decision has not been made for XRPC Procedure endpoints (sent via HTTP POST requests) with regards to support for redirects and therefore should not be relied on.
```

Reworked the wording a little bit. Do we need the word "Unauthenticated" in the above? Can Authenticated XRPC Query's do redirects?

PR #484, opened 2025-12-16
Permission set guide

axfelix:
first draft, rolled up most of the guidance for now


src/app/[locale]/guides/permission-sets/en.mdx
```diff
@@ -0,0 +1,195 @@
+import {Container} from "@/components/Container"
+import clearsky from "./clearsky.png"
+import skyblur from "./skyblur-oauth.png"
+import granular from "./skyblur-granular.png"
+import lexicon from "./skyblur-lexicon.png"
+
+export const metadata = {
+  title: 'Permission Set Guide',
```
Comment (bnewbold) [inline src/app/[locale]/guides/permission-sets/en.mdx]: can we clarify the scope more at the title/link level?

"Permission Requests", description: "A guide for app developers and lexicon designers"

the sets part ("permission sets") is a mechanism folks might know they need to be aware off, so I wouldn't lead with that in link/title/URL

src/app/[locale]/guides/permission-sets/en.mdx
```diff
@@ -0,0 +1,195 @@
...
+
+In general, it is best to use existing permission sets for cleaner UI, but you can also request individual permissions and define your own permission sets as needed. Refer to the [for users](#for-users) section for how these requests appear in the OAuth flow.
+
+### Requesting permissions
+
+When your app initiates an OAuth flow with an ATProto identity provider, it can request specific permission sets by including them in the `scope` parameter of the authorization request. This can be supplied as a *space-separated list* in the `oauth-client-metadata.json` file when registering your OAuth client:
```
Comment (bnewbold) [inline src/app/[locale]/guides/permission-sets/en.mdx]: clients need to declare all the scopes they might ever request in the client metadata document, but then they request the actual scopes for a session during the OAuth flow as part of the authorization request (which is always pushed/PAR for AT OAuth).

usually the scopes are the same for every session; we could maybe skip the concept that the client metadata doc has more than needed. I think this section should mention/clarify that the scope needs to be set in two places; and that if the scope changes it needs to be updated in both places.

src/app/[locale]/guides/permission-sets/en.mdx
```diff
@@ -0,0 +1,195 @@
...
+
+# Permission Sets
+
+## For app developers
+
+Control over AT Proto resources is described and granted using "permissions". A group of a permissions related to a specific Lexicon namespace (record types and API endpoints) can be bundled together as a "permission set". Both are used in the context of [OAuth](/guides/oauth) to grant client software access to account resources on a PDS: For example, the ability to write records of specific types to the user's public repository, or make authenticated API requests to remote services. Developers declare the permissions their app requires to function, and end users are shown the permissions when granting access to the app.
+
+In general, it is best to use existing permission sets for cleaner UI, but you can also request individual permissions and define your own permission sets as needed. Refer to the [for users](#for-users) section for how these requests appear in the OAuth flow.
+
+### Requesting permissions
```
Comment (bnewbold) [inline src/app/[locale]/guides/permission-sets/en.mdx]: I think this section should mention OAuth more explicitly, and link to some high-level OAuth docs entrypoint for client devs

src/app/[locale]/guides/permission-sets/en.mdx
```diff
@@ -0,0 +1,195 @@
...
+When your app initiates an OAuth flow with an ATProto identity provider, it can request specific permission sets by including them in the `scope` parameter of the authorization request. This can be supplied as a *space-separated list* in the `oauth-client-metadata.json` file when registering your OAuth client:
+
+```
+{
+scope: "atproto include:uk.skyblur.permissionSet rpc:app.bsky.actor.getProfile?aud=did:web:api.bsky.app%23bsky_appview repo:app.bsky.feed.post?action=create&action=delete repo:app.bsky.feed.generator?action=create&action=update&action=delete rpc:app.bsky.feed.getFeedGenerator?aud=* rpc:app.bsky.feed.searchPosts?aud=* blob:*/*",
+...
+}
+```
+
+Early AT permission scopes would appear to grant permissions to create, update, or delete any of a user's AT records, including posts from other applications. In general, this should not be necessary — you should try to only request enough permissions to cover the features your app needs. This helps build trust with your users, and minimizes friction during the authorization process.
```
Comment (bnewbold) [inline src/app/[locale]/guides/permission-sets/en.mdx]: I'm not sure what the context of "Early AT permission scopes would appear [...]" is. if you mean the transitional scopes, I don't know if we should even mention here? though might be clarifying. I would just have this paragraph discuss the inherit tension between requesting broader vs narrower scopes.

src/app/[locale]/guides/permission-sets/en.mdx
```diff
@@ -0,0 +1,195 @@
...
+- `identity`: DID and handle
+- `account`: hosting status, email address
+
+Wildcards (*) are allowed in scope string syntax, and grants access to all records — the most common example being `blob:*/*`, for working with images and video. Partial wildcards are *not* supported — for example, `app.bsky.*` is not a valid scope string. If you wanted to request multiple permissions from a particular namespace or Lexicon, you would need to list them out individually, or use a [permission set](#for-lexicon-designers). This is intentional, to avoid overbroad permission requests.
+
+Permission sets are themselves Lexicon schemas and are namespaced like other Lexicons. Unlike other Lexicons, permission sets are prefixed with `include:` in the scope string, and are dynamic; they can be updated.
+
+### Rolling out changes
+
+When you update your app to use new permission sets, it's important to coordinate the software release with the permission-set updates. If you add new permission sets that require user approval, users may need to re-authorize your app to grant these new permissions.
```
Comment (bnewbold) [inline src/app/[locale]/guides/permission-sets/en.mdx]: worth mentioning the client metadata doc again here. for many app devs (eg, mobile apps), the release/deployment process for client metadata will be totally different from the software release process, and they need to be sequenced/coordinated in the right order.

src/app/[locale]/guides/permission-sets/en.mdx
```diff
@@ -0,0 +1,195 @@
...
+
+## For Lexicon designers
+
+Lexicons are authored in JSON — you can find examples in the [Lexicon Guide](/guides/lexicon). Atmosphere HTTP API methods each scope and implement a particular set of Lexicons, so Lexicons must include valid query parameters, request bodies, and response bodies as appropriate. Each Lexicon, in turn, becomes a permission scope — in other words, `app.bsky.feed.post` is an AT record format, a web endpoint, and a granular permission all at once.
```
Comment (bnewbold) [inline src/app/[locale]/guides/permission-sets/en.mdx:72]: I feel like this is a bit "intro to lexicon system in general", but this section is only going to be relevant to people who have already gotten started with lexicons and need to publish a permission-set covering those new resources.

src/app/[locale]/guides/permission-sets/en.mdx
```diff
@@ -0,0 +1,195 @@
...
+        "type": {
+          "description": "Expected values are 'mention' and 'link'.",
+          "type": "string"
+        },
+        "value": {
+          "type": "string"
+...
+```
+
+### Permission Set Design
```
Comment (bnewbold) [inline src/app/[locale]/guides/permission-sets/en.mdx]: I'd probably flip the order and include this content/section above the "how to actually publish with goat" part

src/app/[locale]/guides/permission-sets/en.mdx
```diff
@@ -0,0 +1,195 @@
...
+
+The use of `inheritAud` in the `rpc` permission allows you to [control inheritance behavior](/specs/permission#rpc).
+
+The `blob:*/*` permission for image/video handling cannot be included in permission sets and must always be requested separately.
+
+### Revision
+
+Permission sets can be revised, but removing them or making destructive changes can break existing applications. In general, permission sets should be closely tied to the applications they are designed for, but not necessarily 1:1. Remember that AT Proto applications can utilize multiple different Lexicons.
+
+## For users
```
Comment (bnewbold) [inline src/app/[locale]/guides/permission-sets/en.mdx]: we should also put this info in https://blueskyweb.zendesk.com/hc/en-us.

I'm unsure about mixing actual end user docs in with dev docs, but I do think it is helpful for devs to see what things look and feel like from the perspective of a user. I'd probably re-title / re-frame this section around that.

src/app/[locale]/guides/permission-sets/en.mdx
Comment (axfelix) [reply src/app/[locale]/guides/permission-sets/en.mdx:72]: Yeah, I went back and forth on this, and ultimately I think it's just the inclusion of `goat lex new record` that's redundant with what's already in the docs. Removing that seemed to make the flow too abrupt though, and I want to keep both `lint` and `resolve` in here, so I kept it in and rearranged where this section is located in the guide.

PR #495, opened 2026-02-01
title:langs/detail:langs → title:lang/detail:lang (#491)

bnewbold:
This is re-opening https://github.com/bluesky-social/atproto-website/pull/491 from @mkizka. The previous PR was reverted while we checked the history of the field name, and check what was being used in the network today.

I found several published permission-sets which use `title:lang` and/or `detail:lang`. There are a few permission-sets which define `title:langs` or `detail:langs`, but only as empty objects with no values.

I dug around and couldn't find a specific decision or discussion about changing to the plural version. It's possible that I simply made a mistake when going from the earlier proposals to the final specs.

Given all this, I think we are leaning towards updating the specs, and the indigo implementation, to singular (`title:lang` / `detail:lang`).

Thanks again to @mkizka for catching this before there was too much divergence in the ecosystem!


bnewbold:
(discussed with others and approved)

PR #496, opened 2026-02-02
specs: update many IPLD references to DASL/DRISL references

bnewbold:
The [DASL](https://dasl.ing/) specifications are simpler and align with what we use in atproto.

This updates most references to the IPLD docs to reference the DASL docs instead. It still discusses the lineage in some places, and uses the "CAR v1" language from IPLD (and links to the CAR v1 IPLD specification).

This PR does not update non-English language translations. I think the specs will be at a milestone in a couple months and we should hold off on a translation push until then.

This was a bunch of grepping and copy/paste. I did a superficial self-review skim, but could use a double-check for grammar.

### DIFF
--- a/src/app/[locale]/specs/data-model/en.mdx
+++ b/src/app/[locale]/specs/data-model/en.mdx
@@ -12,11 +12,11 @@ When data needs to be authenticated (signed), referenced (linked by content hash
 
 The schema definition language for atproto is [Lexicon](/specs/lexicon). Other lower-level data structures, like [repository](/specs/repository) internals, are not specified with Lexicons, but use the same data model and encodings.
 
-Distinct pieces of data are called **nodes,** and when encoded in binary (CBOR) result in a **block.** A node may have internal nested structure (maps or lists). Nodes may reference each other by string URLs or URIs, just like with regular JSON on the web. They can also reference each other strongly by hash, referred a **link.** A set of linked nodes can form higher-level data structures like [Merkle Trees](https://en.wikipedia.org/wiki/Merkle_tree) or [Directed Acyclical Graphs (DAG)](https://en.wikipedia.org/wiki/Directed_acyclic_graph). Links can also refer to arbitrary binary data (blobs).
+Distinct pieces of data are called **nodes,** and when encoded in binary (DRISL-CBOR) result in a **block.** A node may have internal nested structure (maps or lists). Nodes may reference each other by string URLs or URIs, just like with regular JSON on the web. They can also reference each other strongly by hash, referred a **link.** A set of linked nodes can form higher-level data structures like [Merkle Trees](https://en.wikipedia.org/wiki/Merkle_tree) or [Directed Acyclical Graphs (DAG)](https://en.wikipedia.org/wiki/Directed_acyclic_graph). Links can also refer to arbitrary binary data (blobs).
 
 Unlike URLs, hash references (links) do not encode a specific network location where the content can be found. The location and access mechanism must be inferred by protocol-level context. Hash references do have the property of being "self-certifying", meaning that returned data can be verified against the link hash. This makes it possible to redistribute content and trust copies even if coming from an untrusted party.
 
-Links are encoded as [Content Identifiers](https://dasl.ing/cid.html) (CIDs), which have both binary and string representations. CIDs include a metadata code which indicates whether it links to a node (CBOR) or arbitrary binary data. Some additional constraints on the use of CIDs in atproto are described below.
+Links are encoded as [Content Identifiers](https://dasl.ing/cid.html) (CIDs), which have both binary and string representations. CIDs include a metadata code which indicates whether it links to a node (DRISL-CBOR) or arbitrary binary data. Some additional constraints on the use of CIDs in atproto are described below.
 
 In atproto, object nodes often include a string field `$type` that specifies their Lexicon schema. Data is mostly self-describing and can be processed in schema-agnostic ways (including decoding and re-encoding), but can not be fully validated without the schema on-hand or known ahead of time.
 
@@ -89,7 +89,7 @@ Note that the legacy format has no `$type` and can only be parsed for known Lexi
 
 atproto uses its own conventions for JSON, instead of using DAG-JSON directly. The main motivation was to have more idiomatic and human-readable representations for `link` and `bytes` in HTTP APIs. The DAG-JSON specification itself mentions that it is primarily oriented toward debugging and development environments, and we found that the use of `/` as a field key was confusing to developers.
 
-Normalizations like key sorting are also not required or enforced when using JSON in atproto: only CBOR is used as a byte-reproducible representation.
+Normalizations like key sorting are also not required or enforced when using JSON in atproto: only DRISL-CBOR is used as a byte-reproducible representation.
 
 The encoding for most of the core and compound types is straight forward, with only `link` and `bytes` needing special treatment.
 
@@ -143,7 +143,7 @@ The use of SHA-256 is a stable requirement in some contexts, such as the reposit
 There are several ways to include a CID hash reference in an atproto object:
 
 - `link` field type (Lexicon type `cid-link`). In DRISL-CBOR, CIDs are encoded by prefixing their binary representation with `0x00`. The CBOR tag used to represent CIDs is 42. In JSON, encodes as `$link` objects (see above).
-- `string` field type, with Lexicon string format `cid`. In CBOR and JSON, encodes as a simple string
+- `string` field type, with Lexicon string format `cid`. In DRISL-CBOR and JSON, encodes as a simple string
 - `string` field type, with Lexicon string format `uri`, with URI scheme `ipld://`
 
 ## Usage and Implementation Guidelines
--- a/src/app/[locale]/specs/event-stream/en.mdx
+++ b/src/app/[locale]/specs/event-stream/en.mdx
@@ -7,7 +7,7 @@ export const metadata = {
 
 # Event Stream
 
-In addition to regular [HTTP API](/specs/xrpc) endpoints, atproto supports continuous event streams. Message schemas and endpoint names are transport-agnostic and defined in [Lexicons](/specs/lexicon). The initial encoding and transport scheme uses binary CBOR encoding over [WebSockets](https://en.wikipedia.org/wiki/WebSocket). {{ className: 'lead' }}
+In addition to regular [HTTP API](/specs/xrpc) endpoints, atproto supports continuous event streams. Message schemas and endpoint names are transport-agnostic and defined in [Lexicons](/specs/lexicon). The initial encoding and transport scheme uses binary [DRISL-CBOR encoding](/lexicons/data-model) over [WebSockets](https://en.wikipedia.org/wiki/WebSocket). {{ className: 'lead' }}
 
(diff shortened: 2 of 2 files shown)


src/app/[locale]/specs/blob/en.mdx
```diff
@@ -19,7 +19,7 @@ The empty blob (zero bytes) is valid in general, though it may be disallowed by
 Currently, the only "blessed" CID type for blobs is similar to that for repository records, but with the `raw` multicodec:
 
 - CIDv1
-- Multibase: binary serialization within DAG-CBOR (or `base32` for JSON mappings)
+- Multibase: binary serialization within CBOR (or `base32` for JSON mappings)
 - Multicodec: `raw` (0x55)
 - Multihash: `sha-256` with 256 bits (0x12)
```
Comment (matthieusieben) [inline src/app/[locale]/specs/blob/en.mdx]: 1) Should we update the naming here to be consistent with https://dasl.ing/cid.html ?
2) Should we mention [DRISL](https://dasl.ing/drisl.html) instead of CBOR, since CBOR in itself does not define tag 42?

```md
- CID version: 1 (0x01)
- Codec: `raw` (0x55)
- Hash type: `sha-256` (0x12) with size of 256 bits (0x20 bytes)

The binary representation of a blob ref CID (as used when serialized into [DRISL](https://dasl.ing/drisl.html)) should always have the following shape: `0x01 0x55 0x12 0x20 <32_bytes_of_the_sha_256_hash>`. Blob ref CIDs can also be encoded into a string representation using (lowercase) base32, prefixed with `b`.
```


Not specifying "JSON mappings" for encoding because CID can be represented as string in other places than `{ $link: "<cid>" }` (lexicons string with `"format": "cid"` like query params, etc.).

src/app/[locale]/specs/data-model/en.mdx
```diff
@@ -131,21 +129,21 @@ For comparison, the DAG-JSON encoding has two nested objects, with outer key `/`
...
+The [original IPFS CID specification](https://github.com/multiformats/cid) is very flexible. It supports a wide variety of hash types, a field indicating the "type" of content being linked to, and various string encoding options. These features are valuable to allow evolution over time, but to maximize interoperability among implementations, only a specific "blessed" set of CID types are allowed. These align with the [DASL CID](https://dasl.ing/cid.html) specification.
 
 The blessed formats for CIDs in atproto are:
 
 - CIDv1
-- multibase: binary serialization within DAG-CBOR `link` fields, and `base32` for string encoding
-- multicodec: `dag-cbor` (0x71) for links to data objects, and `raw` (0x55) for links to blobs
+- multibase: binary serialization within DRISL CBOR `link` fields, and `base32` for string encoding
+- multicodec: `dag-cbor` (0x71) for links to DRISL CBOR data objects, and `raw` (0x55) for links to blobs
```
Comment (matthieusieben) [inline src/app/[locale]/specs/data-model/en.mdx]:
```ms
- CID version: 1 (0x01)
- Codec: `DRISL` (0x71)
- Hash type: `sha-256` (0x12) with size of 256 bits (0x20 bytes)

The binary representation of a CID referencing DRISL data (as used when serialized into [DRISL](https://dasl.ing/drisl.html)) should always have the following shape: `0x01 0x71 0x12 0x20 <32_bytes_of_the_sha_256_hash>`. DRISL CIDs can also be encoded into a string representation using (lowercase) base32, prefixed with `b`.
```

src/app/[locale]/specs/data-model/en.mdx
```diff
@@ -131,21 +129,21 @@ For comparison, the DAG-JSON encoding has two nested objects, with outer key `/`
 
 ## Link and CID Formats
 
-The [IPFS CID specification](https://github.com/multiformats/cid) is very flexible. It supports a wide variety of hash types, a field indicating the "type" of content being linked to, and various string encoding options. These features are valuable to allow evolution over time, but to maximize interoperability among implementations, only a specific "blessed" set of CID types are allowed.
+The [original IPFS CID specification](https://github.com/multiformats/cid) is very flexible. It supports a wide variety of hash types, a field indicating the "type" of content being linked to, and various string encoding options. These features are valuable to allow evolution over time, but to maximize interoperability among implementations, only a specific "blessed" set of CID types are allowed. These align with the [DASL CID](https://dasl.ing/cid.html) specification.
 
 The blessed formats for CIDs in atproto are:
```
Comment (matthieusieben) [inline src/app/[locale]/specs/data-model/en.mdx:134]: We should be more specific here too:

```suggestion
The blessed format for DRISL CIDs (CIDs of DRISL encoded CBOR data) in atproto is:
```

src/app/[locale]/specs/data-model/en.mdx
```diff
@@ -131,21 +129,21 @@ For comparison, the DAG-JSON encoding has two nested objects, with outer key `/`
...
+- multicodec: `dag-cbor` (0x71) for links to DRISL CBOR data objects, and `raw` (0x55) for links to blobs
 - multihash: `sha-256` with 256 bits (0x12) is preferred
 
 The use of SHA-256 is a stable requirement in some contexts, such as the repository MST nodes. In other contexts, like referencing media blobs, there will likely be a set of "blessed" hash types which evolve over time. A balance needs to be struck between protocol flexibility on the one hand (to adopt improved hashes and remove weak ones), and ensuring broad and consistent interoperability throughout an ecosystem of protocol implementations.
 
 There are several ways to include a CID hash reference in an atproto object:
 
-- `link` field type (Lexicon type `cid-link`). In DAG-CBOR encodes as a binary CID (multibase type 0x00) in a bytestring with CBOR tag 42. In JSON, encodes as `$link` object (see above)
-- `string` field type, with Lexicon string format `cid`. In DAG-CBOR and JSON, encodes as a simple string
+- `link` field type (Lexicon type `cid-link`). In DRISL CBOR encodes as a binary CID (multibase type 0x00) in a bytestring with CBOR tag 42. In JSON, encodes as `$link` object (see above)
```
Comment (matthieusieben) [inline src/app/[locale]/specs/data-model/en.mdx]: Do we need to keep the "multibase" work here ? This only makes the spec harder to understand.

```suggestion
- `link` field type (Lexicon type `cid-link`). In DRISL CBOR, CIDs are encoded by prefixing their binary representation with `0x00`. The CBOR tag used to represent CIDs is 42. In JSON, encodes as `$link` objects (see above).
```

src/app/[locale]/specs/data-model/en.mdx
```diff
@@ -8,29 +8,27 @@ export const metadata = {
...
 
-When data needs to be authenticated (signed), referenced (linked by content hash), or stored efficiently, it is encoded in Concise Binary Object Representation (CBOR). CBOR is an IETF standard roughly based on JSON. The specific normalized subset of CBOR used in the atproto data model is called **DAG-CBOR**. All DAG-CBOR data is valid CBOR, and can be read with any CBOR library. Writing or strictly verifying DAG-CBOR with the correct normalization rules sometimes requires additional configuration or a special CBOR implementation. {{ className: 'lead' }}
+When data needs to be authenticated (signed), referenced (linked by content hash), or stored efficiently, it is encoded in Concise Binary Object Representation (CBOR). CBOR is an IETF standard roughly based on JSON. The specific normalized subset of CBOR used in the atproto data model is called [DRISL](https://dasl.ing/drisl.html) (which is very similar to "DAG-CBOR"). All DRISL CBOR data is valid CBOR, and can be read with any CBOR library. Writing or strictly verifying CBOR with the correct normalization rules sometimes requires additional configuration or a special CBOR implementation. {{ className: 'lead' }}
```
Comment (dholms) [inline src/app/[locale]/specs/data-model/en.mdx]: I'd either remove the parenthetical or add a link to the IPLD website for "DAG-CBOR"

src/app/[locale]/specs/data-model/en.mdx
Comment (dholms) [reply src/app/[locale]/specs/data-model/en.mdx:134]: I actually disagree on this one. I think we're specific enough before & after. And referring to "CIDs" here abstractly I think is more clear.

If we do get specific, it should be "DASL CIDs" not "DRISL CIDs"

src/app/[locale]/specs/data-model/en.mdx
Comment (dholms) [reply src/app/[locale]/specs/data-model/en.mdx]: I do like the specificity that Matthieu is getting at better 👍 

Though the Codec can be `DRISL` (`0x71`) or `raw` (`0x55`)

src/app/[locale]/specs/event-stream/en.mdx
```diff
@@ -22,7 +22,7 @@ The initial subscription endpoints are also public and do not require authentica
 
 ## Streaming Wire Protocol (v0)
 
-To summarize, messages are encoded as DAG-CBOR and sent over a binary WebSocket. Clients connect to to a specific HTTP endpoint, with query parameters, then upgrade to WebSocket. Every WebSocket frame contains two DAG-CBOR objects, with bytes concatenated together: a header (indicating message type), and the actual message.
+To summarize, messages are encoded as normalized CBOR ([DRISL](https://dasl.ing/drisl.html))and sent over a binary WebSocket. Clients connect to to a specific HTTP endpoint, with query parameters, then upgrade to WebSocket. Every WebSocket frame contains two CBOR objects, with bytes concatenated together: a header (indicating message type), and the actual message.
```
Comment (dholms) [inline src/app/[locale]/specs/event-stream/en.mdx]: "normalized" -> "deterministic"?

src/app/[locale]/specs/repository/en.mdx
```diff
@@ -96,13 +96,13 @@ When parsing MST data structures, the depth and sort order of keys should be ver
 
 ### CID Formats
 
-The IPFS CID specification is very flexible, and supports a wide variety of hash types, a field indicating the type of content being linked to, and various string encoding options. These features are valuable to allow evolution of the repo format over time, but to maximize interoperability among implementations, only a specific "blessed" set of CID types are allowed.
+The IPFS CID specification is very flexible, and supports a wide variety of hash types, a field indicating the type of content being linked to, and various string encoding options. These features are valuable to allow evolution of the repo format over time, but to maximize interoperability among implementations, only a specific "blessed" set of CID types are allowed. This "blessed" format aligns with the [DASL CID](https://dasl.ing/cid.html) specification.
```
Comment (dholms) [inline src/app/[locale]/specs/repository/en.mdx]: This whole section is pretty redundant with the data-model section. Can we just reference that section rather than repeating?

Otherwise, any decisions we made over in that section, we should also bring over here.

src/app/[locale]/specs/data-model/en.mdx
Comment (matthieusieben) [reply src/app/[locale]/specs/data-model/en.mdx]: If I'm not mistaken, this section is specifically about RAW (this is kinda hard to see in context from Github's PR UI)

Comment (bnewbold) [reply src/app/[locale]/specs/data-model/en.mdx]: this is the data model spec page, not the blob spec page, so both raw and dag-cbor/dasl-cbor makes sense here.

I'll mention both DRISL and `dag-cbor`

src/app/[locale]/specs/event-stream/en.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/event-stream/en.mdx]: went with "DRISL-CBOR". I want to try to be careful with the word pair "deterministic CBOR" because there is a thing called "dCBOR".

src/app/[locale]/specs/repository/en.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/repository/en.mdx]: simplified this a bunch via reference

src/app/[locale]/specs/blob/en.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/blob/en.mdx]: folded most of this in, though kept a separate "Encoding" bullet point instead of adding a line after

PR #501, opened 2026-02-07
feat: add favicon

rdp-studio:
Add favicon.ico so atproto records will look nicer on sites like [PDSls](https://pdsls.dev)


axfelix:
Thanks for clarifying. The favicon used by this website currently is in https://github.com/bluesky-social/atproto-website/blob/main/src/app/%5Blocale%5D/icon.ico — this is a Next.js convention. I don't think adding a redundant file is the right solution here; I'm curious why it isn't being served as a favicon to PDSls and some other platforms.

PR #510, opened 2026-02-15
specs: update repo, sync, and account specs for Sync 1.1

bnewbold:
This finally folds in most of the changes from the Sync 1.1 proposal (https://github.com/bluesky-social/proposals/tree/main/0006-sync-iteration). I also did some general copy-edits and updates to the repo and sync specs.

I compared against daniel's IETF draft (https://www.ietf.org/archive/id/draft-holmgren-at-repository-00.html) when restructuring things, though the level of detail is different enough that copy/paste didn't make sense. I'm open to doing another pass copying over cleaner language and examples from that document in the future though.

These changes should not conflict with (and don't need to block) larger website changes.

This PR does not include:
- updates to semi-spec "guides", like those for account migration and account lifecycle (will hit those separately)
- updates to non-specification docs or tutorials (a lot in flight there)
- much sync motivation or philosophy stuff. eg, don't get in to the concept of "terminating sync"
- a detailed description of what MST nodes inversion requires. I don't think we have a solid technical description of that; the language here matches the original sync proposal and our IETF draft

### DIFF
--- a/src/app/[locale]/specs/account/en.mdx
+++ b/src/app/[locale]/specs/account/en.mdx
@@ -73,7 +73,7 @@ However, in the common case, when there is an active account on a functional cur
 
 - creating an account on the new PDS (which may start with `active` false, such as `deactivated`)
 - migrating data from the old PDS to the new PDS
-- updating identity (DID document) to point to the new PDS and using a new atproto signing key. For PLC DIDs, usually involves updating PLC rotation keys as well. This results in an `#identity` event being emitted from at least one of the PDS hosts.
+- updating identity (DID document) to point to the new PDS and using a new atproto signing key. For PLC DIDs, usually involves updating PLC rotation keys as well. This results in an `#identity` event being emitted from at least the new PDS host, and ideally from the old PDS as well.
 - updating account status on both PDS instances, such that old is inactive, and new is active. This results in `#account` event being emitted from each PDS.
 - emitting a `#sync` and/or `#commit` event from the new PDS, signed by the new atproto signing key, with a higher commit `rev`.
 
--- a/src/app/[locale]/specs/repository/en.mdx
+++ b/src/app/[locale]/specs/repository/en.mdx
@@ -1,12 +1,12 @@
 export const header = {
   title: 'Repository',
   description:
-    'Self-authenticating storage for public account content',
+    'Self-certifying storage for public account content',
 }
 
 # Repository
 
-Each atproto account has a **repository** (or "repo") which stores all of their public data **records**. Repository contents are entirely public and verifiable. Record deletion is supported without leaving a trace or "tombstone" of previous contents. {{ className: 'lead' }}
+Each atproto account has a **repository** (or "repo") which stores all of their public data **records**. Repository contents are entirely public and verifiable ("self-certifying"). Record deletion is supported without leaving a trace or "tombstone" of previous contents. {{ className: 'lead' }}
 
 The repository data structure is a content-addressed [Merkle-tree](https://en.wikipedia.org/wiki/Merkle_tree). Creating, updating, or deleting records (or any other mutations to the repository) changes the root hash value of the overall repository tree. Each published version of the repository tree structure is represented as a **commit**. Commits are cryptographically signed, with rotatable signing keys, which allows recursive authentication of either the repository structure as a whole, or compact "proof chains" for individual records. {{ className: 'lead' }}
 
@@ -149,7 +149,7 @@ Some details about representing repository diffs are "CAR slices":
 - parsing implementation must be tolerant of additional unexpected blocks in the diff, which they should ignore
   - note that unreasonable quantities of unnecessary block data may be considered a form of resource abuse
 
-The diff is a partial Merkle tree, including a signed commit, and can be partially verified. For example, the diff contains a "proof chain" to verify any created or updated records. However, without knowing more about the previous state of the repository, it is not possible to enumerate or confirm which record paths have been deleted.
+The diff is a partial Merkle tree, including a signed commit, and can be partially verified in isolation. For example, the diff contains a "proof chain" to verify any created or updated records. If both metadata about the previous state of the repository and a complete set of record operations are available, then it becomes possible to fully verify the integrity of the diff.
 
 The sync "firehose" mechanism uses a process called "operation inversion" to validate a list of record operations against a diff. In this process, the diff is parsed as a partial MST, and then each operation is applied in reverse: a "create" is applied as a "delete", etc. After all operations have been applied, the root CID (hash) of the MST is recomputed, and should match the `data` field of the previous revision of the repository. If the list of record operations is inaccurate or incomplete, the inverted MST will not match.
 
--- a/src/app/[locale]/specs/sync/en.mdx
+++ b/src/app/[locale]/specs/sync/en.mdx
@@ -17,7 +17,7 @@ As described in the [Repository specification](/specs/repo), each commit to a re
 
 Synchronizing services should reject or ignore repository updates with revision values corresponding to future timestamps (beyond a short fuzzy time drift window). Services can track the commit revision for every account they have seen, and use this to verify synchronization status.
 
-Clients making API requests to a service may want to know if that service has synchronized and indexed recent updates to the most relevant account repository. For example, if a client updates a profile record for an account (by writing to the PDS), and then immediately requests an updated view of the account profile from a separate service, the client would want to know if the record update was applied. Services can indicate synchronization status using the `Atproto-Repo-Rev` HTTP response header, which should contain a single commit revision (TID) of the relevant account. This mechanism is only relevant to API endpoints where a single account was specified.
+Clients making API requests to a service may want to know if that service has synchronized and indexed recent updates to the authenticated account's repository. For example, if a client updates a profile record for their account (by writing to the PDS), and then immediately requests an updated view of their profile from a separate service, the client would want to know if the record update was applied. Services can indicate synchronization status using the `Atproto-Repo-Rev` HTTP response header, which should contain a single commit revision (TID) of the account making the API request.
 
 
 ## Repository Exports
@@ -171,7 +171,7 @@ Here is a summary of validation rules and behaviors:
 - services should verify `#commit` message `blocks` field (CAR diff) against the `ops` list and `prevData` field, using record operation MST inversion
 - services should reject any event messages which exceed reasonable size limits.
 - services should verify that repository data structures are valid against the specification. Missing fields, incorrect MST structure, or other protocol-layer violations should result in events being rejected.
-- services may apply rate-limits to identity, account, commit, and sync events, and suspend accounts or upstream services which violate those limits. Rate limits might also be applied to recovery modes such as invalid signatures resulting in an identity refresh, missing or out-of-order commits, etc.
+- services may apply rate-limits to identity, account, commit, and sync events, and throttle accounts or upstream services which violate those limits. Rate limits might also be applied to recovery modes such as invalid signatures resulting in an identity refresh, missing or out-of-order commits, etc.
 - services should ignore `#commit` and `#sync` events with a `rev` lower or equal to the most recent successfully processed `rev` for that DID, and should reject commit events with a `rev` corresponding to a future timestamp (beyond a clock drift window of a few minutes)
 - services should check the `since` value in `#commit` events, and if it is not consistent with the most recent `rev` for that DID, mark the repo as out-of-sync
 - similarly, services should check `#commit` message `prevData` values against the most recent commit object `data` field


src/app/[locale]/specs/account/en.mdx
```diff
@@ -20,14 +20,16 @@ Identity metadata (DID documents and handle status), and the account hosting sta
 
 Network services are encouraged to implement API endpoints (such as `com.atproto.sync.getRepoStatus`) which describe current hosting status for individual accounts. If they expose an event stream, they should also emit `#account` events when their local account hosting status updates for an account.
 
-In addition to the `active` boolean, an account might in a more specific state (all of which correspond with `active=false`):
+In addition to the `active` boolean, an account might in a more specific state:
```
Comment (dholms) [inline src/app/[locale]/specs/account/en.mdx]:
```suggestion
In addition to the `active` boolean, an account might be in a more specific state:
```

src/app/[locale]/specs/account/en.mdx
```diff
@@ -71,9 +73,9 @@ However, in the common case, when there is an active account on a functional cur
 
 - creating an account on the new PDS (which may start with `active` false, such as `deactivated`)
 - migrating data from the old PDS to the new PDS
-- updating identity (DID document) to point to the new PDS and using a new atproto signing key. For PLC DIDs, usually involves updating PLC rotation keys as well.
-- updating account status on both PDS instances, such that old is inactive, and new is active
-- emitting a `#commit` event from the new PDS, signed by the new atproto signing key, with a higher commit `rev`.
+- updating identity (DID document) to point to the new PDS and using a new atproto signing key. For PLC DIDs, usually involves updating PLC rotation keys as well. This results in an `#identity` event being emitted from at least one of the PDS hosts.
```
Comment (dholms) [inline src/app/[locale]/specs/account/en.mdx]: I think this should always be emitted by the new PDS. And preferably emitted by the old PDS as well.

src/app/[locale]/specs/repository/en.mdx
```diff
@@ -6,40 +6,37 @@ export const header = {
...
 
-Repositories and their contents are canonically stored in [DRISL](https://dasl.ing/drisl.html) CBOR, as a graph of data objects referencing each other by content hash (CID Links). Large binary blobs are not stored directly in repositories, though they are referenced by hash ([CID](https://github.com/multiformats/cid)). This includes images and other media objects. Repositories can be exported as [CAR](https://dasl.ing) files for offline backup, account migration, or other purposes. {{ className: 'lead' }}
+In the atproto network architecture, the authoritative location of an account's repository is the associated Personal Data Server (PDS). An account's current PDS location is declared in the DID Document. {{ className: 'lead' }}
```
Comment (dholms) [inline src/app/[locale]/specs/repository/en.mdx:15]: I think it's worth adding a note about "self-certifying" in here. The description of the page is "Self-authenticating storage for public account content", but we don't actually talk about that at all in the introduction. (also maybe change authenticated -> certifying in the page description)

src/app/[locale]/specs/repository/en.mdx
```diff
@@ -109,59 +106,63 @@ The standard file format for storing data objects is Content Addressable aRchive
...
+- the root CID indicated in the CAR header (the first element of `roots`) must point to the new commit block (which must be included)
+- all MST nodes in the current repo revision which did not exist in the previous repo revision must be included
+- any additional MST nodes needed to support "operation inversion" must be included (described below)
+- all required blocks must be included even if they appeared previously in the history of the repository.
+  - for example, if a record is created in rev C, deleted in rev F, and re-created in rev N, the diff "since F" must include the record block
+- all "created" records must be included
+- any records which have been "deleted" should not have the record value included
+- any records which have been "updated" should include the new value, and should not include the previous version
+- if a record value has been deleted from one path and created (or updated) at a new path in the same diff (eg, was "moved"), it should be included
+- parsing implementation must be tolerant of additional unexpected blocks in the diff, which they should ignore
```
Comment (dholms) [inline src/app/[locale]/specs/repository/en.mdx:149]: We might need to clarify how this plays with the "pre-order deterministic CAR file".

One option is that we just say deterministic CARs don't matter for diffs (which honestly I think is probably fine)

src/app/[locale]/specs/repository/en.mdx
```diff
@@ -109,59 +106,63 @@ The standard file format for storing data objects is Content Addressable aRchive
...
+  - note that unreasonable quantities of unnecessary block data may be considered a form of resource abuse
 
-However, an observer which does not know the full state of the repository at the "older" revision *can not* reliably enumerate all of the records that have been removed from the repository. Such an observer also can not see the previous values of deleted or updated records, either as full values or by CID. Note that the later is an intentional design goal for the diff concept: it is desired that content deletion happen rapidly and not "draw attention" to the content which has been deleted. It is technically possible for "archival" observers to track deletion events and lookup the previous content value, but this requires additional resources and effort.
+The diff is a partial Merkle tree, including a signed commit, and can be partially verified. For example, the diff contains a "proof chain" to verify any created or updated records. However, without knowing more about the previous state of the repository, it is not possible to enumerate or confirm which record paths have been deleted.
```
Comment (dholms) [inline src/app/[locale]/specs/repository/en.mdx]: I feel like this is a little bit confusing ahead of the next paragraph.

We say "it's not possible" & then say "the firehose relies on it". Maybe just calling me attention to "the diff in isolation is not enough". Or "the diff requires additional metadata about the repo & operations in order to validate them"

src/app/[locale]/specs/repository/en.mdx
```diff
@@ -109,59 +106,63 @@ The standard file format for storing data objects is Content Addressable aRchive
...
-- if a host receives a “since” request earlier than the oldest available revision for a repository, it should return the full repository. This may happen if the host does not have the complete history of the repository.
-    - for example, if a repository had revisions "140", "150", and "160", then migrated to a new PDS and revisions continued "161" and "170", if the new PDS is asked for a diff "since" 150, the new PDS would probably need to return the full repository, because the earliest revision it would be aware of was "160" or "161" (depending on how migration was implemented).
+In some cases the inversion process requires the inclusion of contextual MST nodes in the diff, which would not otherwise need to be included. For example, MST nodes that reference record paths directly adjacent (in sorted order) to those mutated in the diff. The definition of which nodes need to be included in the diff is ultimately tautological: those MST nodes which are necessary for the inversion process. A set of test vectors is included in the [atproto test case git repository](https://github.com/bluesky-social/atproto-interop-tests).
```
Comment (dholms) [inline src/app/[locale]/specs/repository/en.mdx:156]: Didn't we have another definition of this? I guess we never actually proved it...

src/app/[locale]/specs/sync/en.mdx
```diff
@@ -6,31 +6,44 @@ export const header = {
 
 # Data Synchronization
 
-One of the main design goals of atproto (the "Authenticated Transfer Protocol") is to reliably distribute public content between independent network services. This data transfer should be trustworthy (cryptographicly authenticated) and relatively low-latency even at large scale. It is also important that new participants can join the network at any time and “backfill” prior content.
+Synchronization of public data between independent network participants is one of the defining features of the atproto (which is named for "Authenticated Transfer"). Data redistibution is trustworthy, low-latency, and resource efficient at large scale. {{ className: 'lead' }}
```
Comment (dholms) [inline src/app/[locale]/specs/sync/en.mdx]:
```suggestion
Synchronization of public data between independent network participants is one of the defining features of atproto (which is named for "Authenticated Transfer"). Data redistibution is trustworthy, low-latency, and resource efficient at large scale. {{ className: 'lead' }}
```

src/app/[locale]/specs/sync/en.mdx
```diff
@@ -6,31 +6,44 @@ export const header = {
...
 
-This section describes the major data synchronization features in atproto. The primary real-time data synchronization mechanism is repository event streams, commonly referred to as "firehoses". The primary batch data transfer mechanism is repository exports as CAR files. These two mechanisms can be combined in a "bootstrapping" process which result in a live-synchronized copy of the network.
+There are two main data synchronization mechanisms in atproto. Batch data transfer is supported using full-repository exports as CAR files over HTTP. Real-time synchronization can be provided by a repository event stream (over WebSocket), commonly referred to as a "firehose". Used together, they make it possible to bootstrap a complete, live-updated, and authenticated copy of network data. {{ className: 'lead' }}
```
Comment (dholms) [inline src/app/[locale]/specs/sync/en.mdx]: HTTPS?

Comment (dholms) [reply src/app/[locale]/specs/sync/en.mdx]:
```suggestion
There are two main data synchronization mechanisms in atproto. Batch data transfer is supported using full-repository exports as CAR files over HTTP. Real-time synchronization can be provided by a repository event stream (over WebSocket), commonly referred to as a "firehose". Used together, they enable a complete, live-updated, and authenticated copy of network data. {{ className: 'lead' }}
```

src/app/[locale]/specs/sync/en.mdx
```diff
@@ -6,31 +6,44 @@ export const header = {
...
+As described in the [Repository specification](/specs/repo), each commit to a repository has a *revision* number, encoding as a [TID](/specs/tid) string. Revision numbers are used a logical clock to track synchronization status individual repositories. The revision number must always increase between commits for the same repository, even if the account migrates between hosts or has an extended period of inactivity.To simplify revision management, PDS implementations should use the current wall time (converted to TID) as the revision for every commit. When using wall time, it is still important to check that the new revision is higher than the previous revision, to account for clock drift or concurrent updates.
```
Comment (dholms) [inline src/app/[locale]/specs/sync/en.mdx]:
```suggestion
As described in the [Repository specification](/specs/repo), each commit to a repository has a *revision* value, encoding as a [TID](/specs/tid) string. Revisions function as a logical clock to track synchronization status individual repositories. The revision must always increase between commits for the same repository, even if the account migrates between hosts or has an extended period of inactivity. To simplify revision management, PDS implementations should use the current wall time (converted to TID) as the revision for every commit. When using wall time, it is still important to check that the new revision is higher than the previous revision, to account for clock drift or concurrent updates.
```

We use "revision" or "revision value" in IETF draft. I don't really think of these as "numbers"

src/app/[locale]/specs/sync/en.mdx
```diff
@@ -6,31 +6,44 @@ export const header = {
...
 
-## Firehose
+Synchronizing servics should reject or ignore repository updates with revision numbers corresponding to future timestamps (beyond a short fuzzy time drift window). Services can track the commit revision for every account they have seen, and use this to verify synchronization status.
```
Comment (dholms) [inline src/app/[locale]/specs/sync/en.mdx]:
```suggestion
Synchronizing services should reject or ignore repository updates with revision values corresponding to future timestamps (beyond a short fuzzy time drift window). Services can track the commit revision for every account they have seen, and use this to verify synchronization status.
```

src/app/[locale]/specs/sync/en.mdx
```diff
@@ -6,31 +6,44 @@ export const header = {
...
+Clients making API requests to a service may want to know if that service has synchronized and indexed recent updates to the most relevant account repository. For example, if a client updates a profile record for an account (by writing to the PDS), and then immediately requests an updated view of the account profile from a separate service, the client would want to know if the record update was applied. Services can indicate synchronization status using the `Atproto-Repo-Rev` HTTP response header, which should contain a single commit revision (TID) of the relevant account. This mechanism is only relevant to API endpoints where a single account was specified.
```
Comment (dholms) [inline src/app/[locale]/specs/sync/en.mdx]: Right now this actually always returns the _calling users_ DID (not the specified DID in the call).

src/app/[locale]/specs/sync/en.mdx
```diff
@@ -6,31 +6,44 @@ export const header = {
...
 
-In most cases the repository data synchronized over the firehose is self-certifying (contains verifiable signatures), and consumers can verify content without making additional requests directly to account PDS instances. It is possible for services to redact events from the firehose, such that downstream services would not be aware of new content.
 
-Identity and account information is *not* self-certifying, and services may need need to verify independently. This usually means independent DID and [handle resolution](/specs/handle). Account hosting status might also be checked at account PDS hosts, to disambiguate hosting status at different pieces of infrastructure.
+## Repository Exports
 
-The event message types are declared in the `com.atproto.sync.subscribeRepos` Lexicon schema, and are summarized below. A few fields are the same for all event types (except for `repo` vs `did` for `#commit` events):
+Full repository exports can be fetch from the account's PDS host using the `com.atproto.sync.getRepo` XRPC endpoint (HTTP GET). This endpoint is not authenticated, and returns all repo records, MST nodes, and the current signed commit object, all in a single CAR file.
```
Comment (dholms) [inline src/app/[locale]/specs/sync/en.mdx]:
```suggestion
Full repository exports can be fetched from the account's PDS host using the `com.atproto.sync.getRepo` XRPC endpoint (HTTP GET). This endpoint is not authenticated, and returns all repo records, MST nodes, and the current signed commit object, all in a single CAR file.
```

src/app/[locale]/specs/sync/en.mdx
```diff
@@ -6,31 +6,44 @@ export const header = {
...
+
+Multiple network services provide firehose endpoints under the same `com.atproto.sync.subscribeRepos` stream endpoint, with compatible message types and semantics. PDS hosts provide a firehose that includes updates for all hosted accounts. **Relays** are network services which subscribe to multiple upstream firehoses (eg, multiple PDS hosts) and aggregate them in to a single combined event stream. A relay which attempts to aggregate accounts from all PDS instances in the network outputs a "full-network" firehose.
+
+Repository data synchronized over a firehose is self-certifying and contains verifiable signatures. Consuming services can verify synchronized data without making additional requests to the account's PDS host. Missing updates to a single repository detectable as gaps in the stream of `#commit` messages. However, if an intermediary service were to filter all messages pertaining to repository, that would not necessarily be detectable by consuming services.
```
Comment (dholms) [inline src/app/[locale]/specs/sync/en.mdx]:
```suggestion
Repository data synchronized over a firehose is self-certifying and contains verifiable signatures. Consuming services can verify synchronized data without making additional requests to the account's PDS host. Missing updates to a single repository are detectable as gaps in the stream of `#commit` messages. However, if an intermediary service were to filter all messages pertaining to repository, that would not necessarily be detectable by consuming services.
```

Comment (dholms) [reply src/app/[locale]/specs/sync/en.mdx]: also last sentence here reads a bit funny to me

"However, an intermediary service could completely remove all messages pertaining to a specific repository without being detected"?

Or maybe even just X-ing this sentence

src/app/[locale]/specs/sync/en.mdx
```diff
@@ -6,31 +6,44 @@ export const header = {
...
 
 - `seq` (integer, required): used to ensure reliable consumption, as described in Event Streams
-- `did` / `repo`(string with DID syntax, required): the account/identity associated with the event
-- `time` (string with datetime syntax, required): an informal and non-authoritative estimate of when event was received. Intermediate services may decide to pass this field through as-is, or update to the current time
+- `did` (string with DID syntax, required): the account associated with the event. The `#commit` message is inconsistent and uses `repo` as the field name
+- `time` (string with datetime syntax, required): an informal and non-authoritative estimate of when event was received. Intermediary services may decide to pass this field through as-is, or update to the current time
+
+Firehose event stream messages have a hard maximum size limit of 5 MBytes, measured as WebSocket frames. This this is inclusive of all encoding and nesting overhead, and rules out some messages which would not otherwise exceed lexicon schema limits. Messages are generally expected to be well below this hard size limit. This limit may evolve over time.
```
Comment (dholms) [inline src/app/[locale]/specs/sync/en.mdx]:
```suggestion
Firehose event stream messages have a hard maximum size limit of 5 MBytes, measured as WebSocket frames. This is inclusive of all encoding and nesting overhead, and rules out some messages which would not otherwise exceed lexicon schema limits. Messages are generally expected to be well below this hard size limit. This limit may evolve over time.
```

Comment (dholms) [reply src/app/[locale]/specs/sync/en.mdx]: Also `#commit` messages are limited to 2MBytes by the lexicon. That seems more relevant to call out than the 5 MB limit (I'm actually not sure where the 5MB limit comes from?)

src/app/[locale]/specs/sync/en.mdx
```diff
@@ -89,80 +100,130 @@ Event Fields:
...
+
+If the chain of `#commit` messages is found to have broken, or a `#sync` message indicates that the repository state has changed, then the service will need to re-synchronize the repository. This often means fetching the full repo CAR export.
+
+If many services attempt to re-synchronize a repository at the same time, the upstream PDS host may be overwhelmed with a "thundering herd" of requests. To mitigate this, receiving services should first attempt to fetch the repo CAR file from their direct upstream (often a relay instance). That gives the upstream an opportunity to coalesce and cache the repository export, which distributes load. The upstream may instead simply use an HTTP redirect to the PDS instance.
+
+The "Record-Level Synchronization" second below describes a design pattern for tracking record state and handling re-synchronization incidents.
```
Comment (dholms) [inline src/app/[locale]/specs/sync/en.mdx]:
```suggestion
The "Record-Level Synchronization" section below describes a design pattern for tracking record state and handling re-synchronization incidents.
```

src/app/[locale]/specs/sync/en.mdx
```diff
@@ -89,80 +100,130 @@ Event Fields:
...
-- data limits on records specifically should be verified. Events containing corrupt or entirely invalid records may be rejected. for example, a record not being CBOR at all, or exceeding normal data size limits.
-- more subtle data validation of records may be enforced, or may be ignored, depending on the service. For example, unsupported CID hash types embedded in records should probably be ignored by Relays (even if they violate the atproto data model), but may result in the record or commit event being rejected by an AppView
-- mirroring services, which retain a full copy of repository data, should verify that commit diffs leave the MST tree in a complete and valid state (eg, no missing records, no invalid MST nodes, commit CID would be reproducible if the MST structure was re-generated from scratch)
-- Relays (specifically) should not validate records against Lexicons
+- services may apply rate-limits to identity, account, commit, and sync events, and suspend accounts or upstream services which violate those limits. Rate limits might also be applied to recovery modes such as invalid signatures resulting in an identity refresh, missing or out-of-order commits, etc.
```
Comment (dholms) [inline src/app/[locale]/specs/sync/en.mdx]: throttle in stead of suspend? or maybe "throttle or suspend"

src/app/[locale]/specs/repository/en.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/repository/en.mdx:15]: added to intro paragraph

src/app/[locale]/specs/repository/en.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/repository/en.mdx:149]: I'd lean towards holding off on formally specifying it one way or the other until we've finished implementations

src/app/[locale]/specs/repository/en.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/repository/en.mdx:156]: I don't think we have a more formal definition, but I might be forgetting something?

PR #526, opened 2026-02-18
fix mobile

syui:
The json view overflows on mobile.


axfelix:
Can't duplicate issue, but the fix seems fine — merging!

PR #537, opened 2026-02-18
Add screen reader text for navigation items

richardkrikler:


src/components/Header.tsx
```diff
@@ -104,9 +104,11 @@ export const Header = forwardRef<
           <ul role="list" className="flex items-center gap-6">
             <TopLevelNavItem href="https://bsky.app/profile/atproto.com">
               <BlueskyIcon className="h-5 w-5 fill-slate-900 dark:fill-slate-50" />
+              <span className='sr-only'>AT Protocol on Bluesky</span>
             </TopLevelNavItem>
             <TopLevelNavItem href="https://github.com/bluesky-social/atproto">
               <GithubIcon className="h-6 w-6 fill-slate-900 dark:fill-slate-50" />
+              <span className='sr-only'>AT Protocol on GitHub</span>
             </TopLevelNavItem>
```
nulfrost: I think since these are links it makes more sense to use an ``aria-label`` here instead of screenreader only text https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA8

richardkrikler: Both solutions are viable options.
In this repo both variants are already used... (e.g. in Footer.tsx)

Footer.tsx:
```tsx
<Link href={href} className="group">
  <span className="sr-only">{children}</span>
  <Icon className="h-5 w-5 fill-zinc-700 transition group-hover:fill-zinc-900 dark:group-hover:fill-zinc-500" />
</Link>
```

Header.tsx:
```tsx
<Link href="/" aria-label="Home">
  <Logo className="h-6" />
</Link>        
```

PR #557, opened 2026-02-24
Update Japanese translations

yamarten:
This update includes both overall improvements to the existing Japanese translation and follow-up to the original text update.


axfelix:
Builds fine and looks good!

PR #561, opened 2026-02-25
list more community SDKs

axfelix:
This is probably overdue! Somewhat concerned I've potentially selected one among many equally complete Swift solutions, most of the others were relatively more mature than others for that language. Want to avoid picking winners but great to get more coverage here imo.

### DIFF
--- a/src/components/Libraries.tsx
+++ b/src/components/Libraries.tsx
@@ -82,7 +82,7 @@ const official: LibraryGroupProp[] = [
           serviceAuth: 'full',
           crypto: 'full',
           repo: 'full',
-          plc: 'none',
+          plc: 'full',
           oauthServer: 'none',
         },
       },
@@ -143,8 +143,8 @@ const community: LibraryGroupProp[] = [
           identifiers: 'full',
           bskyHelpers: 'partial',
           lexicon: 'full',
-          identity: 'partial',
-          streaming: 'partial',
+          identity: 'full',
+          streaming: 'full',
           serviceAuth: 'partial',
           crypto: 'full',
           repo: 'full',
@@ -233,14 +233,14 @@ const community: LibraryGroupProp[] = [
         description: 'Created by @mackuba.eu.',
         status: {
           identifiers: 'partial',
-          bskyHelpers: 'partial',
-          lexicon: 'none',
+          bskyHelpers: 'none',
+          lexicon: 'partial',
           identity: 'full',
           streaming: 'full',
-          serviceAuth: 'partial',
+          serviceAuth: 'none',
           crypto: 'none',
           repo: 'none',
-          plc: 'none',
+          plc: 'partial',
           oauthServer: 'none',
         },
       },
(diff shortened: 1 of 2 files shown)


src/components/Libraries.tsx
```diff
@@ -90,6 +98,158 @@ const community: LibraryProp[] = [
...
+      bsky: 'none',
+      crypto: 'none',
+      mst: 'partial',
+      lexicon: 'partial',
+      identity: 'partial',
+      streaming: 'none',
+      serviceAuth: 'none',
+      plc: 'none',
+      oauthServer: 'partial',
+    },
```
Ovyerus: Are there any (official) guidelines on what constitutes a "full" vs "partial" status? Would love to know what exactly I can improve upon in some of these areas because I personally feel like some of them (identity, lexicon, httpClient) would be considered fully done.

Also small nitpick, atex currently doesn't have *anything* to deal with MSTs just yet :P

src/components/Libraries.tsx
```diff
@@ -90,6 +98,158 @@ const community: LibraryProp[] = [
...
+      bsky: 'partial',
+      crypto: 'none',
+      mst: 'partial',
+      lexicon: 'none',
+      identity: 'partial',
+      streaming: 'none',
+      serviceAuth: 'partial',
+      plc: 'none',
+      oauthServer: 'none',
+    },
```
mackuba: Hey - you can link to https://ruby.sdk.blue instead, and then you can also add "streaming" (Skyfall) and I think count "identity" as "full" (DIDKit) 😉

mackuba: Also, can a link to https://sdk.blue fit somewhere there below the list for people looking for other languages? :>

src/components/Libraries.tsx
```diff
@@ -16,239 +16,321 @@ import logoSwift from '@/images/logos/swift.svg'
...
+        description: 'Created by @nonbinary.computer.',
+        status: {
+          identifiers: 'full',
+          bskyHelpers: 'partial',
+          lexicon: 'full',
+          identity: 'partial',
+          streaming: 'partial',
+          serviceAuth: 'partial',
+          crypto: 'none',
+          repo: 'none',
```
orual: appreciate adding jacquard. should note that it has a repo implementation, used by tranquil PDS. interested in what gap in completeness for service auth might be, similar for streaming support. has all crypto needed for repo commit signing and validation.

axfelix: Good call! Sorry, I'm being conservative with this feature matrix and expect to get feedback that I've been too conservative. Probably need to add an explanation of completeness at the bottom of the page :)

orual: Yeah some clarification would be good. I think one of you guys had a list of "things needed for SDK completeness" kicking around in an older version of the docs or a GitHub discussion, but I'd need to track it down, and there were a few things that I remember being a bit unclear on that front.

axfelix: Yup, you're probably thinking of https://github.com/bluesky-social/atproto/discussions/2415!

Anyway, take a look at most recent commit: https://github.com/bluesky-social/atproto-website/pull/561/changes/1b854c7ba33ce192b6eaba165fc4a3418bb5a598

orual: Ah yep that was it. And cool, thanks. I would bump up jacquard's streaming support to 'complete' (supports all extant streaming endpoints and they're pretty well tested now), and same for 'identity'. PLC ops and OAuth server support are the main actual gaps, presently, and it will never get more than 'partial' Bluesky support as a deliberate design decision. It can do preference ops via a generic system, as well as Bluesky rich text, and handle the client side of moderation label processing, but it's not a Bluesky SDK, it's an atproto SDK, and deliberately avoids privileging Bluesky uses more than it must for a good developer experience.

mackuba: For the Ruby SDK I would change:

- plc -> partial (`didkit` - lookup of did documents & audit logs)
- serviceAuth -> none I think
- lexicon -> partial if that includes "XRPC client"
- bsky helpers -> none

src/components/Libraries.tsx
axfelix: Toggled these for now (actually removed httpClient as a standalone pill as you'll see) and added an explanation toward the bottom of the page!

PR #574, opened 2026-03-06
update SDK patterns

axfelix:
Catching up with some recent SDK changes eg https://github.com/bluesky-social/cookbook/pull/54

### DIFF
--- a/src/app/[locale]/guides/sdk-auth/en.mdx
+++ b/src/app/[locale]/guides/sdk-auth/en.mdx
@@ -51,8 +51,6 @@ If you are using password auth, you should generate an [app password](/specs/xrp
         "",                          // authToken (optional 2FA token)
         nil,                         // refresh callback (optional)
       )
-
-      fmt.Printf("Logged in as: %s\n", client.AccountDID)
     }
     ```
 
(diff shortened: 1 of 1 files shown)


src/app/[locale]/guides/sdk-auth/en.mdx
```diff
@@ -21,42 +21,36 @@ If you are using password auth, you should generate an [app password](/specs/xrp
...
 
-      client := &xrpc.Client{
-        Host: "https://bsky.social", // or your PDS host
-      }
-
-      session, _ := comatproto.ServerCreateSession(ctx, client, &comatproto.ServerCreateSession_Input{
-        Identifier: "your-handle.bsky.social",
-        Password:   "your-app-password",
-      })
+      client, _ := atclient.LoginWithPasswordHost(ctx,
```
Comment (bnewbold) [inline src/app/[locale]/guides/sdk-auth/en.mdx:47]: I think it is worth doing `atclient.LoginWithPassword` here, instead of supplying the PDS hostname. Needing to look up the hostname as another step is a pain and something folks will skip.

For examples like this, can do:

```
dir := identity.DefaultDirectory()
```

and pass that as an arg; or even just instantiate DefaultDirectory as the arg itself.

Comment (axfelix) [reply src/app/[locale]/guides/sdk-auth/en.mdx:47]: Sounds good, will update! Honestly I've mostly guessed at the intended Go *client* patterns and just verified they work since nobody has been super vocal about Go client dev.

PR #594, opened 2026-03-19
label guide: note the 'bot' self-label value

bnewbold:
This is sort of documenting a bsky-specific label in the context of the overall protocol, which isn't great. But this is a place devs will look for documentation, and I do expect/assume this would be a value others would use in other apps.


axfelix (approved):
Good catch, I think this is fairly innocuous in terms of BSky-centrism and pairs well with https://github.com/bluesky-social/atproto-website/pull/593

PR #616, opened 2026-04-12
updates to XRPC service auth spec (aud and iss fields)

bnewbold:
This PR corresponds to the "revised" proposal for XRPC service auth JWTs. AKA, the "single-string" syntax, not "separate fields" syntax.

This also clarifies that the `lxm` field is required for XRPC service auth (not optional). 

This PR does not include updates to the DID spec page to clarify the syntax for fragment syntax (for service endpoint identifiers, and public keys).

Note that this rolls up the `iss` change along with the `aud` change. We might switch and have `iss` be two-string format (with `kid` indicating the key identifier), or just not touch that in this iteration.


src/app/[locale]/specs/xrpc/en.mdx
```diff
@@ -123,6 +123,12 @@ The PDS acts as a generic proxy between clients and other atproto services. Clie
 
 The HTTP header is `atproto-proxy`, and the value is a DID (identifying a service), followed by a service endpoint identifier, joined by a `#` character. The PDS resolves the service DID, extracts a service endpoint URL from the DID document, and proxies the request on to the identified server.
 
+<Note>
+Until Spring 2026, the reference PDS implementation did not pass through the full service reference to the `aud` field of the JWT auth header when proxying requests. It would strip the service endpoint identifier, and only include the bare DID.
+
+This is in the process of being updated to include both the DID and service identifier fragment.
```
Comment (axfelix) [inline src/app/[locale]/specs/xrpc/en.mdx:130]: Should we implement this before merging so we aren't stuck with "in the process of being updated"? Otherwise this looks good.

src/app/[locale]/specs/xrpc/en.mdx
```diff
@@ -93,21 +93,22 @@ The JWT parameters are:
     - use `ES256K` for `k256` keys
     - use `ES256` for `p256` keys
 - `typ` header field (string, required): currently `JWT`, but intend to update to a more specific value.
-- `iss` body field (string, required): account DID that the request is being sent on behalf of. This may include a suffix service identifier; see below
-- `aud` body field (string, required): service DID associated with the service that the request is being sent to
+- `kid` header field (string, optional): issuer key identifier, used to specify which public key was used to sign the JWT. Corresponds to identifier fragment in DID document, specifying which "verification method" was used. This field should not include the DID or the hash separator (`#`). Defaults to `atproto`. Note that this field is in the (unsigned) JWT header, not the body. See below for semantics and verification guidance.
```
Comment (matthieusieben) [inline src/app/[locale]/specs/xrpc/en.mdx]: I would prefer to use the DID URL syntax here:

can be either:
 - relative (to issuer) `#my-key`
 - absolute `did:plc:123#my-key` (with a DID that MUST be equal to the issuer)

defaulting to `#atproto`.

If we don't want absolute DID-URLs, I would still keep the `#` prefix as it keeps the syntax more consistent with the DID spec. It also makes the syntax distinct from what is typically used in oauth land, which should help avoiding confusion as to where to find this key when encountering such tokens.

Comment (matthieusieben) [reply src/app/[locale]/specs/xrpc/en.mdx]:
```suggestion
- `kid` header field (string, optional): issuer key identifier, used to specify which public key was used to sign the JWT. Corresponds to identifier fragment in DID document, specifying which "verification method" was used. This field should not include the DID or the hash separator (`#`). Defaults to `atproto`. See below for semantics and verification guidance.
```

src/app/[locale]/specs/xrpc/en.mdx
```diff
@@ -93,21 +93,22 @@ The JWT parameters are:
...
+- `iss` body field (string, required): account DID ("issuer") that the request is being sent on behalf of.
+- `aud` body field (string, required): service DID associated with the service that the request is being sent to ("audience"). May include an optional fragment name identifying the specific service type and endpoint. See [DID syntax specification](/specs/did) for syntax details, and below for semantics. New software should always include the service fragment name when generating JWTs.
```
Comment (matthieusieben) [inline src/app/[locale]/specs/xrpc/en.mdx]:
```suggestion
- `aud` body field (string, required): service DID associated with the service that the request is being sent to ("audience"). Should include a fragment name identifying the specific service type and endpoint. New software should always include the service fragment name when generating JWTs. Legacy services may accept `aud` values without fragment. See [DID syntax specification](/specs/did) for syntax details, and below for semantics.
```

src/app/[locale]/specs/xrpc/en.mdx
```diff
@@ -93,21 +93,22 @@ The JWT parameters are:
...
 - JWT signature (string, required): base64url-encoded signature using the account DID's signing key.
 
-When the token is generated in the context of a specific service in the issuer's DID document, the issuer field may have the corresponding *service* identifier in the `iss` field, separated by a `#` character. For example, `did:web:label.example.com#atproto_labeler` for a labeler service. When this is included the appropriate signing key is determined based on a fixed mapping of service identifiers to key identifiers:
+The issuer field (`iss`) is the DID of the account sending the request. If an issuer key identifier (`kid`) field is provided, it indicates a specific public key from the account's DID document in the `verificationMethod` array, defaulting to `atproto`. Note that `iss` is a JWT body field, while `kid` is a JWT header field. Receiving services must only accept specific key types (verification method identifiers), to maintain security separation between registered keys. This usually means accepting only `atproto`, unless there is a specific application-specific or service-specific additional key type expected (eg, `atproto_label`).
```
Comment (matthieusieben) [inline src/app/[locale]/specs/xrpc/en.mdx]: Add something like:

> Recipients must always check that the key referenced by the `kid` (if present) matches a verification method identifier present in the DID Document of the issuer (`iss`).

src/app/[locale]/specs/xrpc/en.mdx
```diff
@@ -93,21 +93,22 @@ The JWT parameters are:
...
 - `exp` body field (number, required): token expiration time, as a UNIX timestamp with seconds precision. Should be a short time window, as revocation is not implemented. 60 seconds is a good token lifetime.
 - `iat` body field (number, required): token creation time, as a UNIX timestamp with seconds precision.
-- `lxm` body field (string, optional): short for "lexicon method". NSID syntax. Indicates the endpoint that this token authorizes. Servers must always validate this field if included, and should require it for security-sensitive operations. May become required in the future.
+- `lxm` body field (string, optional): short for "lexicon method". NSID syntax. Indicates the endpoint that this token authorizes. This field is required for all authenticated XRPC endpoint requests.
 - `jti` body field (string, required): unique random string nonce. May be used by receiving services to prevent reuse of token and replay attacks.
```
Comment (matthieusieben) [inline src/app/[locale]/specs/xrpc/en.mdx]:
```suggestion
- `jti` body field (string, required): unique random string nonce. Should be used by receiving services to prevent reuse of token and replay attacks.
```

matthieusieben (commented):
Looking good !

src/app/[locale]/specs/xrpc/en.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/xrpc/en.mdx:130]: There is going to need to be a transition period between when we publish this update (and update our services/SDKs), and when everybody else in the ecosystem (eg, feed gens and appviews) do the updates on their end.

I'll lay this out in the (forthcoming) updated proposal.

src/app/[locale]/specs/xrpc/en.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/xrpc/en.mdx]: I tweaked the paragraph a bit to make this clearer.

src/app/[locale]/specs/xrpc/en.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/xrpc/en.mdx]: I'm against having the `iss` DID repeated in this field. It feels redundant/inefficient, and implementation complexity (another thing to verify).

I don't have super strong feelings about whether to include the `#`, but would lean towards not? If feels unexpected.

I might be open to other syntax to differentiate this field from other `kid` types? But keeping it simple and just being the identifier value feels like the best move.

Open to discussing further.

Comment (bnewbold) [reply src/app/[locale]/specs/xrpc/en.mdx]: we decided to go ahead with `#` prefix in the `kid` field, but not the full DID.

src/app/[locale]/specs/xrpc/en.mdx
Comment (bnewbold) [reply src/app/[locale]/specs/xrpc/en.mdx:130]: Here is that full proposal, including transition plan: https://github.com/bluesky-social/proposals/tree/main/0014-service-auth-revised

bnewbold:
Updated to clarify the `#` prefix syntax. From my perspective this is ready to merge now, but will wait for an explicit approval.

matthieusieben (approved):
Nice work @bnewbold ! This will finally allow us to `getServiceAuth` using OAuth credentials.

src/app/[locale]/specs/xrpc/en.mdx
Comment (mackuba) [reply src/app/[locale]/specs/xrpc/en.mdx:130]: I think the "Until Spring 2026, the reference PDS implementation did not pass through" part is slightly misleading, because AFAICT it still doesn't pass through?…

Comment (axfelix) [reply src/app/[locale]/specs/xrpc/en.mdx:130]: Fair observation, @bnewbold I think the "in the process of getting updated" is still not ideal language here

Comment (bnewbold) [reply src/app/[locale]/specs/xrpc/en.mdx:130]: could we move this to a dedicated issue? I have some thoughts

PR #618, opened 2026-04-13
docs: add progressive scope upgrade patterns to scopes guide

zzstoatzz:
## summary

adds a "Requesting scopes progressively" section to the [scopes guide](https://atproto.com/guides/scopes), covering:

- **selecting scopes before authorization**: client-side pattern where users choose permissions before the OAuth redirect, building the scope string dynamically from selections
- **upgrading scopes for an existing session**: server-side (BFF) pattern where the app stores a preference, initiates a new OAuth flow with expanded scopes, and replaces the session on callback
- **permission set dynamism**: when updating a published set is sufficient vs when an explicit re-authorization is required (cross-namespace integrations)

this felt like a natural extension of the existing scopes guide content ("which scopes to request") — it addresses what happens when the answer changes over time for a given user.

## examples from the wild

- [pds.ls](https://github.com/notjuliet/pdsls/blob/main/src/auth/scope-flow.ts)
- [plyr.fm](https://github.com/zzstoatzz/plyr.fm/blob/main/backend/src/backend/api/auth.py#L601-L635)

### DIFF
--- a/src/app/[locale]/guides/scopes/en.mdx
+++ b/src/app/[locale]/guides/scopes/en.mdx
@@ -41,6 +41,62 @@ The auth flow page will display all the relevent permissions being granted. This
 
 Client Apps should be sure to check which auth scopes were actually approved during the auth flow. In particular, when requesting permission to read an account email address (`account:email`), the user might decline that permission while granting others. In theory, any individual permission might be denied while the overall request is granted.
 
+## Requesting scopes progressively
+
+An application does not always need all of its permissions at sign-in. Using broad scopes like `transition:generic` is the quickest way to get a prototype working, but it grants the app access to nearly everything in the user's account — and the authorization dialog reflects that, which can discourage people from trying it.
+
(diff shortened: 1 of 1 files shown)


axfelix:
Thanks Nate! Will take a look at this later today.

zzstoatzz:
sorry meant to open as a draft! still need to QA

zzstoatzz:
ok, i decided this fold this into the existing scopes guide, and trim it down a bit

happy to move this around, address any nits, close if too in-the-weeds, or hand it off for copy/wording changes! just let me know 🫡

zzstoatzz:
note i saw this: https://underreacted.leaflet.pub/3mjfozhlhys2z which is awesome but felt like color that could be added to other pages, whereas here i'm trying to speak to

"well how do i expand scopes when a configured experience needs more scopes than the default experience without going full `transition:generic`?"

axfelix:
OK! I'm actually about to refactor a couple pages in our Auth docs to make them flow better, but right now I'd like to just merge this contribution as-is so that it doesn't fall out of the refactor. Thanks for the help!

PR #622, opened 2026-04-14
Consolidate blog scripts under a single entry point

jimray:
All blog-related commands now run through `npm run blog <command>`:
create, remove, ssite, hide-reply, create-publication.


axfelix (approved):
Looks fine to me! good refactor

PR #625, opened 2026-04-14
Blog index standard site

jimray:
Steve Simkins's blog post

### DIFF
--- a/src/app/[locale]/blog/indexing-standard-site/en.mdx
+++ b/src/app/[locale]/blog/indexing-standard-site/en.mdx
@@ -39,7 +39,7 @@ The problem that eventually killed it was bandwidth. What I hadn't fully account
 
 ## What worked: Jetstream \+ Cloudflare
 
-The fix was switching data sources rather than rearchitecing aroudn Tap. [Jetstream](https://docs.bsky.app/blog/jetstream) is a lighter WEbSocket service that does the collecting filtering upstream — it delivers only the collections you subscribe to — no local filtering of the full firehose. The tradeoff is there's no built-in database, and no backfilling but for docs.surf, which only needs the latest 100 posts, that's fine.
+The fix was switching data sources rather than rearchitecing aroudn Tap. [Jetstream](https://docs.bsky.app/blog/jetstream) is a lighter WebSocket service that does the collecting filtering upstream — it delivers only the collections you subscribe to — no local filtering of the full firehose. The tradeoff is there's no built-in database and no backfill but for docs.surf, which only needs the latest 100 posts, that's fine.
 
 The architecture that's been stable and cheap:
 
--- a/src/app/[locale]/blog/indexing-standard-site/en.mdx
+++ b/src/app/[locale]/blog/indexing-standard-site/en.mdx
@@ -39,7 +39,7 @@ The problem that eventually killed it was bandwidth. What I hadn't fully account
 
 ## What worked: Jetstream \+ Cloudflare
 
-[Jetstream](https://docs.bsky.app/blog/jetstream) solved the bandwidth problem. Unlike Tap, it delivers only the collections you subscribe to — no local filtering of the full firehose. The tradeoff: no built-in database, and no backfilling. For docs.surf, which only needs the latest 100 posts, that's fine.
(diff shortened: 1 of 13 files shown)


src/app/[locale]/blog/indexing-standard-site/en.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+Queuing the resolution work via Cloudflare helped. Tap's webhook support made this straightforward: when a valid event comes in, Tap posts a payload to a Cloudflare Worker, which drops it onto a queue for async processing. That worked well enough that docs.surf launched on this architecture.
+
+The problem that eventually killed it was bandwidth. What I hadn't fully accounted for is that Tap consumes the entire firehose and filters locally — it's receiving every event on the network, not just the collection you care about. Running it on Railway, egress costs climbed whenever Standard.site adoption ticked up. I briefly moved the instance to a home server to cut costs, and promptly got throttled by my ISP from the incoming bandwidth alone. The firehose is not a small stream.
+
+## What worked: Jetstream \+ Cloudflare
+
+[Jetstream](https://docs.bsky.app/blog/jetstream) solved the bandwidth problem. Unlike Tap, it delivers only the collections you subscribe to — no local filtering of the full firehose. The tradeoff: no built-in database, and no backfilling. For docs.surf, which only needs the latest 100 posts, that's fine.
```
Comment (axfelix) [inline src/app/[locale]/blog/indexing-standard-site/en.mdx]: I'm not sure we intend to contrast Tap with Jetstream 1:1 like this — since I think of Tap in some ways as just a websocket client with some other features, and Jetstream is an alternative, more minimal websocket. Might reframe this a bit to "just connecting to the Jetstream websocket directly" to be clearer about it (and, honestly, I think Tap should support this method of operation in the future to be cleaner about it, hopefully this case study is motivating, cc @dholms).

src/app/[locale]/blog/indexing-standard-site/en.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+* A **Cloudflare Durable Object** maintains the Jetstream WebSocket connection, keeping all traffic within Cloudflare's network
+* Incoming records are batched and sent to a **Cloudflare Queue** for async processing
+* The queue worker handles the multi-step resolution and verification, with retry logic for the verification timing issue (more on that below)
+* Resolved documents land in **Cloudflare D1**
+* A cron job re-checks records that initially failed verification
+
+Total cost: $5/month.
+
+One thing to know about publishing `standard.site` records: there's a race condition baked into the publishing flow. To publish a post, you create the atproto record first, get the AT URI, then deploy your site with the appropriate `<link>` tag. There's an inevitable gap between record creation and site deployment — so if your indexer tries to verify immediately after seeing the record, it will get a false negative. The cron-based re-verification pass is the fix; don't try to handle this synchronously.
```
Comment (axfelix) [inline src/app/[locale]/blog/indexing-standard-site/en.mdx:54]: Similar to the above comment about how Tap should probably be improved to handle this use case without workarounds, can we speculate here about how we want this to work in the future without needing a cron pass? Is it just durable queues, do we want to standardize those more in this ecosystem?

axfelix (approved):
This looks good! Left a couple comments about potential followups, not sure they need to block the blog post.

src/app/[locale]/blog/indexing-standard-site/en.mdx
Comment (jimray) [reply src/app/[locale]/blog/indexing-standard-site/en.mdx]: Good call, clarified that

src/app/[locale]/blog/indexing-standard-site/en.mdx
Comment (jimray) [reply src/app/[locale]/blog/indexing-standard-site/en.mdx:54]: Yeah, and this one gets tricke, from my experience with standard site. There's actually another loop, that I didn't want to get into, which is that many standard site indexers *will* try to verify immediately so the fix is for publishers to publish their `standard.site` record again with an updated timestamp after the post with the verification URI is published.

I do think that's a question for the standard site crew

Comment (axfelix) [reply src/app/[locale]/blog/indexing-standard-site/en.mdx:54]: My last employer was essentially in the "make every job its own durable queue rather than use cron schedules" business so I am pretty familiar with the problem space here, I just wonder if this is on the list of "things that shouldn't be left as implementation details"

PR #626, opened 2026-04-15
Auth refactor

axfelix:


src/app/[locale]/guides/permission-requests/en.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+Permissions relate to user owned resources on PDS instances are are represented by five types:
+
+- `repo`: Public Repository (records and collections)
+- `rpc`: Service Authentication (API calls to external services)
+- `blob`: uploaded media files
+- `identity`: DID and handle
+- `account`: hosting status, email address
+
+Wildcards (*) are allowed in scope string syntax, and grants access to all records — the most common example being `blob:*/*`, for working with images and video. Partial wildcards are *not* supported — for example, `app.bsky.*` is not a valid scope string. If you wanted to request multiple permissions from a particular namespace or Lexicon, you would need to list them out individually, or use a [permission set](/guides/permission-sets). This is intentional, to avoid overbroad permission requests.
```
Comment (gaearon) [inline src/app/[locale]/guides/permission-requests/en.mdx]: > Partial wildcards are *not* supported — for example, `app.bsky.*` is not a valid scope string.

you probably mean `repo: app.bsky.*` here

src/app/[locale]/guides/permission-requests/en.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+You should try to only request enough permissions to cover the features your app needs. We provide [transition scopes](/specs/oauth#authorization-scopes) that grant broad permissions to create, update, or delete any of a user's AT records, including posts from other applications — but in general, these broad permission requests should be avoided. Granular permissions help build trust with your users and minimize friction during the authorization process.
+
+From a user experience perspective, it's best to keep the number of requested permission sets low. Each additional permission set adds more information to the authorization dialog, which can lead to decision fatigue. It is possible to allow users to grant only a subset of requested permissions, but this can lead to edge cases.
+
+## Permission types
+
+Permissions relate to user owned resources on PDS instances are are represented by five types:
+
+- `repo`: Public Repository (records and collections)
```
Comment (gaearon) [inline src/app/[locale]/guides/permission-requests/en.mdx]: how do you feel about a single inline example per each. just to orient the reader

src/app/[locale]/guides/permission-requests/en.mdx
```diff
@@ -0,0 +1,127 @@
...
+    'A permissions guide for app developers and Lexicon designers',
+}
+
+This page explains how to **use** permission scopes in your OAuth flow. If you're building an app with custom Lexicons and want to **define and publish** your own permission sets, see [Permission Sets](/guides/permission-sets).
+
+## Overview
+
+This guide describes how to request and define permissions for your atproto application. [OAuth](/guides/auth) handles authorization, and this page covers how to specify *what* your app can access.
+
+Control over atproto resources is granted using [permissions](/specs/permission). Individual permissions like `repo:com.example.post` grant access to specific record types or API endpoints. Related permissions can be bundled into a **permission set** with a human-readable title and description — these are what users see in the authorization dialog.
```
Comment (gaearon) [inline src/app/[locale]/guides/permission-requests/en.mdx]: > to specific record types

maybe “to writing specific record types”?

it’s easy to misunderstand permissions as having to do with reads

src/app/[locale]/guides/permission-requests/en.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+- `repo`: Public Repository (records and collections)
+- `rpc`: Service Authentication (API calls to external services)
+- `blob`: uploaded media files
+- `identity`: DID and handle
+- `account`: hosting status, email address
+
+Wildcards (*) are allowed in scope string syntax, and grants access to all records — the most common example being `blob:*/*`, for working with images and video. Partial wildcards are *not* supported — for example, `app.bsky.*` is not a valid scope string. If you wanted to request multiple permissions from a particular namespace or Lexicon, you would need to list them out individually, or use a [permission set](/guides/permission-sets). This is intentional, to avoid overbroad permission requests.
+
+Permission sets are themselves Lexicon schemas and are namespaced like other Lexicons. Unlike other Lexicons, permission sets are prefixed with `include:` in the scope string, and are dynamic; they can be updated.
```
Comment (gaearon) [inline src/app/[locale]/guides/permission-requests/en.mdx]: hmm i’m not sure i follow the “unlike other lexicons” here.. they’re always the same syntax in the scope string right? `type:lexicon?options`.

i think here too a single concrete example would drive the point better than describing it in text

src/app/[locale]/guides/permission-sets/en.mdx
```diff
@@ -1,69 +1,59 @@
...
-When you update your app to use new permission sets, it's important to coordinate the software release with the permission-set updates. If you add new permission sets that require user approval, users will need to re-authorize your app to grant these new permissions.
+### Minimal permissions
 
-Your release and deployment process for the [client metadata document](/specs/oauth#clients) may be separate from your app's code deployment process. Make sure to plan accordingly — an updated `oauth-client-metadata.json` that requests new permissions must be live before a version of your app that requires those permissions.
+The absolute minimum for an app using atproto OAuth (identity-only, no data access):
+- `atproto`
 
-## For Lexicon designers
+If you also need Bluesky profile data:
+- `rpc:app.bsky.actor.getProfile?aud=did:web:api.bsky.app#bsky_appview`
```
Comment (gaearon) [inline src/app/[locale]/guides/permission-sets/en.mdx:51]: you can already read bluesky profile from records. maybe worth mentioning that and then saying “but if you want to hit the Bluesky API with user’s authentication for (something the records don’t give?)” then do this

src/app/[locale]/guides/permission-requests/en.mdx
Comment (axfelix) [reply src/app/[locale]/guides/permission-requests/en.mdx]: Yup, good clarification here!

src/app/[locale]/guides/permission-requests/en.mdx
Comment (axfelix) [reply src/app/[locale]/guides/permission-requests/en.mdx]: Feeling good about it, turned this into a table 👍

src/app/[locale]/guides/permission-sets/en.mdx
Comment (axfelix) [reply src/app/[locale]/guides/permission-sets/en.mdx:51]: It's a good point, but I think qualifying it further is a bit more confusing, and I've seen people stating in the past that they needed this permission for an expected flow to work, so I'm inclined to leave it for now.

src/app/[locale]/guides/permission-requests/en.mdx
Comment (axfelix) [reply src/app/[locale]/guides/permission-requests/en.mdx]: Rewrote this slightly (the "unlike" part only really applies to the update clause), good catch.

src/app/[locale]/guides/about-oauth/en.mdx
```diff
@@ -0,0 +1,87 @@
...
+- [OAuth with Go](/guides/go-oauth-cli-tutorial) - OAuth from a Go CLI app
+- [OAuth with Next.js](/guides/oauth-tutorial) - Build a web app with AT Protocol login
+
+For a conceptual overview of how OAuth works in atproto (including how it differs from other OAuth systems), see the [Auth overview](/guides/auth#what-is-o-auth).
+
+## SDK Libraries
+
+The following libraries handle the OAuth flow, including [DPoP](/guides/oauth-patterns#understanding-d-po-p), token refresh, and session management.
+
+### TypeScript
```
Comment (jimray) [inline src/app/[locale]/guides/about-oauth/en.mdx:21]: Maybe include atcute: https://npmx.dev/package/@atcute/oauth-browser-client

src/app/[locale]/guides/about-oauth/en.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+- [@atproto/oauth-client-browser](https://github.com/bluesky-social/atproto/tree/main/packages/oauth/oauth-client-browser#readme) ([npmjs](https://www.npmjs.com/package/@atproto/oauth-client-browser)) - Suitable for frontend-only applications (e.g. SPAs).
```
Comment (jimray) [inline src/app/[locale]/guides/about-oauth/en.mdx]: What about linking to npmx.dev instead of npmjs ?

src/app/[locale]/guides/auth/en.mdx
```diff
@@ -56,25 +56,14 @@ OAuth aims to solve these problems with a family of open specifications for mana
 
 As a developer, you can request limited access to a person's account, meaning if you only need to post a message to someone's public timeline, you don't also need to request access to read their private messages. This is great for your app because you don't have to worry about managing data you don't need, and it's great for the person using your app because they don't need to worry about their DMs being compromised.
 
-## How is OAuth Different in AT Protocol?
```
Comment (jimray) [inline src/app/[locale]/guides/auth/en.mdx:59]: I kind of miss having this section here. I see it in the OAuth patterns doc, that feels a little late.

src/app/[locale]/guides/oauth-patterns/en.mdx
```diff
@@ -1,58 +1,23 @@
...
+- **Migration**: Atproto users can migrate their accounts between servers ([PDSes](/guides/glossary#pds-personal-data-server)) over time. To facilitate this atproto has a flexible [Identity](/guides/identity) layer, which allows usernames ([handles](/specs/handle)) to be resolved to a static user ID ([DID](/specs/did)), which in turn can be resolved to locate the user's PDS. When a user logs in to an app, the OAuth client dynamically resolves these relationships.
```
Comment (jimray) [inline src/app/[locale]/guides/oauth-patterns/en.mdx:18]: It feels like this could be split into two points: that atproto users can migrate their accounts *and* the oauth client dynamically resolves the account. The first is a feature of atproto that's definitely relevant to developers, the second is a feature for oauth clients

src/app/[locale]/guides/oauth-patterns/en.mdx
```diff
@@ -1,58 +1,23 @@
...
-- [Client Assertion Backend](https://github.com/bluesky-social/proposals/tree/main/0010-client-assertion-backend): A proposed alternative to the TMB pattern with simplified server-side logic.
-
-Of these three, the BFF pattern is the currently recommended approach for building confidential-client applications. This is demonstrated in the [cookbook/go-oauth-web-app](https://github.com/bluesky-social/cookbook/tree/main/go-oauth-web-app#readme) and [cookbook/python-oauth-web-app](https://github.com/bluesky-social/cookbook/tree/main/python-oauth-web-app#readme) examples.
+- **Client IDs**: In other OAuth ecosystems, it is often necessary for client apps to pre-register themselves with the resource server. This is not viable in a decentralized system (with many clients and many resource servers), so atproto addresses this using [Client ID Metadata Documents](/specs/oauth#client-id-metadata-document).
```
Comment (jimray) [inline src/app/[locale]/guides/oauth-patterns/en.mdx]: I think we could include some more details about the Client ID, even something like: "Your client_id is a URL, which is fetched at auth time to determine OAuth features like redirect URI and scopes."

If we wanted to expand more, we could so something like "In a centralized OAuth call, the client_id is a unique public identifier for an app that is registered with the centralized provider. This isn't viable for decentralized apps, so the client_id is a URL (unique by definition) that gets fetched at authorization time and provides details like the redirect URI and scopes the app is requesting"

src/app/[locale]/guides/oauth-patterns/en.mdx
```diff
@@ -1,58 +1,23 @@
...
 
 ## Understanding DPoP
```
Comment (jimray) [inline src/app/[locale]/guides/oauth-patterns/en.mdx:22]: We don't mention PKCE or PAR anywhere other than the spec, really. And while those are typically handled by SDKs, they're also gotchas if you're coming from a centralized OAuth system. It might make sense to expand this section to "Understanding the decentralized OAuth security stack: PKCE, PAR, and DPoP"

PKCE protects the authorization code *in transit*
PAR protects the authorization itself (params never appear in a URL)
DPoP protects the token once it's been issued.

and include links to the specs. I don't think most devs will implement these from scratch, but they are new concepts if you're coming from centralized OAuth land.

src/app/[locale]/guides/oauth-patterns/en.mdx
```diff
@@ -68,11 +33,89 @@ DPoP is mandatory in atproto OAuth. **If you're using an SDK**, it's handled aut
...
+
+// start a new OAuth flow with the expanded scope
+const url = await oauthClient.authorize(handle, {
+  scope: wantsIntegration ? EXTENDED_SCOPE : BASE_SCOPE,
+});
+```
+
+The server stores the old session identifier alongside the OAuth state so the callback handler knows to replace it. On subsequent logins, the server can look up the user's stored preferences and include the expanded scopes automatically.
+
+Note that [permission sets](/specs/permission#permission-sets) can reduce the need for explicit scope upgrades: when a client refreshes its tokens, the computed permissions may be updated to reflect changes to the resolved sets. If a new capability falls within an existing permission set's [namespace authority](/specs/permission#namespace-authority), updating the published set is sufficient — no re-authorization required. Cross-namespace integrations will still require an explicit upgrade. See [Rolling out changes](/guides/permission-requests#rolling-out-changes) for guidance on coordinating scope changes with application releases.
```
Comment (jimray) [inline src/app/[locale]/guides/oauth-patterns/en.mdx:122]: I feel like we can do more to nudge/push people towards permission sets? Would moving this to a section after "progressive scope requests" do it?

src/app/[locale]/guides/permission-requests/en.mdx
```diff
@@ -0,0 +1,133 @@
...
+
+Permission sets are themselves Lexicon schemas and are namespaced like other Lexicons. Permission sets are prefixed with `include:` in the scope string. Unlike other Lexicons, they can be updated in [most contexts](#rolling-out-changes).
+
+## Clients
+
+OAuth client software is identified by a globally unique `client_id`. Distinct variants of client software may have distinct `client_id` values; for example the browser app and Android (mobile OS) variants of the same software might have different `client_id` values. The `client_id` must be a fully-qualified web URL from which the client-metadata JSON document can be fetched. For example, `https://app.example.com/oauth-client-metadata.json`. Some more about the `client_id`:
```
Comment (jimray) [inline src/app/[locale]/guides/permission-requests/en.mdx:64]: I think it's worth pointing out that the `client_id` being a fully qualified URL is unique (I think?) to atproto, or at least distributed oauth. Usually it's just a hash.

src/app/[locale]/guides/permission-requests/en.mdx
```diff
@@ -0,0 +1,133 @@
...
+
+OAuth in AT uses [Pushed Authorization Requests](https://auth0.com/blog/what-are-oauth-push-authorization-requests-par/) (PAR). The client metadata file will be fetched dynamically during the session lifecycle, and the `scope` parameter in `oauth-client-metadata.json` *needs to match* the actual authorization HTTP request made by your app:
+
+<OauthAuthUrl />
+
+You should try to only request enough permissions to cover the features your app needs. We provide [transition scopes](/specs/oauth#authorization-scopes) that grant broad permissions to create, update, or delete any of a user's AT records, including posts from other applications — but in general, these broad permission requests should be avoided. Granular permissions help build trust with your users and minimize friction during the authorization process.
+
+From a user experience perspective, it's best to keep the number of requested permission sets low. Each additional permission set adds more information to the authorization dialog, which can lead to decision fatigue. It is possible to allow users to grant only a subset of requested permissions, but this can lead to edge cases.
+
+## Permission types
```
Comment (jimray) [inline src/app/[locale]/guides/permission-requests/en.mdx:43]: I think we can do more to explain the difference between `repo` and `rpc` here.

Maybe an example like "Request `repo` when you need to write or edit specific data from a user's PDS. Request `rpc` when you need access to a service, such as an endpoint that collates posts into a feed"

Or maybe frame it as `repo` gives access to the PDS, `rpc` gives access to the AppView?

src/components/Navigation.tsx
```diff
@@ -280,9 +280,10 @@ export const navigation: Array<NavGroup> = [
         href: '/guides/auth',
         links: [
           { title: 'SDK authentication', href: '/guides/sdk-auth' },
-          { title: 'OAuth patterns', href: '/guides/oauth-patterns' },
-          { title: 'Scopes', href: '/guides/scopes' },
-          { title: 'Permission Requests', href: '/guides/permission-sets' },
+          { title: 'About OAuth', href: '/guides/about-oauth' },
+          { title: 'Permission Requests', href: '/guides/permission-requests' },
```
jimray, line 284: I think we should use the word `scope` in the nav, since that's a thing people will be looking for

Maybe `Permission: Requesting Scopes`
and
`Permission Sets: Bundling Scopes`

jimray (commented):
Oh man this is great. I don't think there's anything that's blocking in my feedback, mostly drilling into the differences with decentralized auth

src/components/Navigation.tsx
axfelix: Believe it or not I changed this in response to feedback that people were primarily looking for the word "Permission", and we have limited space in the nav. Only compromise I can think of is something hideous like "Scopes / Permission Sets" which I can live with.

src/app/[locale]/guides/permission-requests/en.mdx
Comment (axfelix) [reply src/app/[locale]/guides/permission-requests/en.mdx:43]: Hmmmm. That actually feels too simplifying to me rather than v/v — I like that we break it down as "one is for literally writing records, the other is for calls to endpoints."

Repository != PDS and AppView != App endpoint, so I'm wary of changing this.

src/app/[locale]/guides/permission-requests/en.mdx
Comment (axfelix) [reply src/app/[locale]/guides/permission-requests/en.mdx:64]: Client ID Metadata documents are in general semi-unique to how we're doing things (it's an IETF standard that Emelia is working on, I think we're the de facto adopters). I might just rather add a link to this spec anchor (https://atproto.com/specs/oauth#client-id-metadata-document) from this paragraph rather than go into more detail here, does that work?

src/app/[locale]/guides/oauth-patterns/en.mdx
Comment (axfelix) [reply src/app/[locale]/guides/oauth-patterns/en.mdx:122]: OAuth patterns is now the *last* doc in the section that rolls up a bunch of miscellaneous advice rather than introducing concepts — which makes the mention of permission sets here more of a back reference. I think it's OK for a back reference to read as more of a reminder at this point (I peppered in a *lot* of additional links to sets elsewhere).

src/app/[locale]/guides/auth/en.mdx
Comment (axfelix) [reply src/app/[locale]/guides/auth/en.mdx:59]: I could be convinced to put it back — the top level page was feeling long, whereas this made a really nice intro to the grab bag of a patterns page.

src/app/[locale]/guides/oauth-patterns/en.mdx
Comment (axfelix) [reply src/app/[locale]/guides/oauth-patterns/en.mdx:18]: Hm. I see it, but as written, this is basically glancing at the first point to establish the second, which is what I'd expect the framing *in the auth docs* to do.

PR #628, opened 2026-04-16
OAuth widget

jimray:
Added a widget for testing OAuth permission sets and creating OAuth scope strings

- **chore: add vitest for unit testing**
- **feat(scope-builder): add shared TypeScript types**
- **feat(scope-builder): encodeAudDid percent-encodes DID service fragment**
- **feat(scope-builder): NSID and partial-wildcard validators**
- **feat(scope-builder): buildScopeString renders single permissions**
- **feat(scope-builder): assembleScopeString with dedupe and atproto-first invariant**
- **feat(scope-builder): include: string, namespace check, and permission-set JSON builder**
- **feat(scope-builder): curated data for permission sets and individual scopes**
- **feat(scope-builder): scope-builder custom element with permission sets, individual scopes, and output panel**
- **feat(scope-builder): permission-author custom element with progressive scope/set output**
- **feat(scope-builder): client-side React loader and type declarations**
- **feat(scope-builder): add /guides/scope-builder MDX page with widget embed**
- **feat(scope-builder): add Scope Builder to the Auth sidebar navigation**
- **fix(scope-builder): remove runtime import of .d.ts file**
- **feat(scope-builder): split into independent scope-builder and permission-author loaders**
- **feat(scope-builder): use real published Bluesky permission sets**
- **feat(scope-builder): single-column layout with sticky scope string**
- **feat(scope-builder): Permission Set Builder layout and UX improvements**
- **docs(scope-builder): reorganize guide page with both widgets**
- **Updated error location**


axfelix:
OK! Looking at this properly now:

In general, looks great, I think it's very nice to be able to offer anything interactive like this, it will help make creating permission scopes way more tangible. Nitpicks as follows:

- The "collection NSID" field in the "Add permission" dialog would really benefit from having a find-as-you-type search for at least the contents of the atproto-lexicons and bsky-lexicons repositories.
- The "permission set metadata" dialog feels even more open ended in a way that's intimidating — I wonder if adding find as you type to the other dialog will be illustrative here.
- The way that permissions eventually get displayed as just space-separated rather than newline-separated in the output is technically correct to our standard but can be really hard to read — I wonder if we should just format it with newlines (but have it copy and paste out with spaces) for legibility in the widget.
- The Bluesky permission set list winds up just being kinda long in terms of how it's displayed on the page, and I found myself unsure why I was scrolling through all these Bluesky permissions in particular while trying to get to the end of it. We should find a way to collapse this further I think (eg maybe putting the descriptions behind an accordion like the bundled permissions).

jimray:
Good callouts!

3 and 4 are easy fixes, pushing those soon.

If we fix 1, it should go a long way towards fixing 2. Those lexicons are pretty static, so we can probably just bundle those? We'll want to make sure we have something in place for new lexicons down the line.

jimray:
After a bit of a think, I want to go back to the purpose of the permission set widget: to encourage devs to build their own permission sets for their own apps. I agree the open endedness is intimidating, but want to focus on the issue we want to solve here, and I think it might be more straightforward:

* Better placeholder text with a plausible third-party NSID that teaches by example — `com.acme.myapp.authBasic` instead of `com.example.myApp.permissions`.
* A subtle "see an example" link under the NSID field that opens a real-world permission set (maybe [chat.bsky.authFullChatClient](https://pds.ls/at://did:plc:4v4y5r3lwsbtmsxhile2ljac/com.atproto.lexicon.schema/chat.bsky.authFullChatClient) or even a developer one like Beacon Bits just added on PDSls) to show what the finished product looks like.
* A one-line help text under the NSID field: "Use your app's reverse-domain namespace (e.g., com.yourapp.authFeatureName)."

I think we can probably add a few more examples and some screenshots in the [permissions sets](https://atproto.com/guides/permission-sets) docs to help with the "why".

src/components/ScopeBuilder/scopeUtils.ts
```diff
@@ -0,0 +1,222 @@
...
+  if (candidateNsid === '*') return true
+  const setParts = setNsid.split('.')
+  const setNamespace = setParts.slice(0, -1).join('.')
+  if (!setNamespace) return false
+  return (
+    candidateNsid === setNamespace ||
+    candidateNsid.startsWith(setNamespace + '.')
+  )
+}
+
```
jimray, line 147: @matthieusieben IncludeScope.isParentAuthorityOf (and isAllowedPermission) are protected, so this is reimplemented it by hand. Does that seem ok to you?

PR #671, opened 2026-06-05
Off Protocol schedule widget

jimray:
- **Added blueskyPostUrl to codetv post**
- **Add next-livestream widget to Off Protocol page**


axfelix:
Nice work!! Looking good in the preview.

PR #681, opened 2026-06-09
move to next 16

axfelix:


axfelix:
fyi @jimray this is kind of ugly but I wanted dependabot off my back, looks fine everywhere I can tell

jimray:
I'm sure it'll be great!

PR #704, opened 2026-06-29
did:plc key terminology

yamarten:
["Recovery Keys" section](https://atproto.com/guides/account-recovery#recovery-keys) contains several statements that may be confusing or inconsistent with [the current `did:plc` specification](https://web.plc.directory/spec/v0.1/did-plc). This PR updates the terminology and descriptions accordingly.

* "Recovery key" is an [older](https://github.com/bluesky-social/atproto-website/blob/060f6a4b6b401c92f1e539ccb6c988000ce7387f/content/specs/did-plc.md#signing-and-recovery-keys) name for what is now called a rotation key. My understanding is that the term is still sometimes used for an emergency rotation key controlled by the user, but there does not seem to be a reason to use a different term only in this section.
* DID documents no longer include rotation keys. They are still public, but they are stored in the operation log.
* The signing key cannot modify the DID document. A PDS holds a rotation key in order to update the DID document, but the repository signing key is separate.
* The 72-hour recovery window applies to nullifying PLC operations signed by lower-priority rotation keys, not directly to overriding the repository signing key.

As an alternative, this section could explain the model by separating the three roles more explicitly: the repository signing key, the PDS-controlled rotation key, and the user-controlled recovery/rotation key.

### DIFF
--- a/src/app/[locale]/guides/account-recovery/en.mdx
+++ b/src/app/[locale]/guides/account-recovery/en.mdx
@@ -26,14 +26,14 @@ Most atproto accounts use `did:plc`, a DID method that relies on **rotation keys
 
 By adding your own rotation key with higher priority than your PDS's key, you gain the ability to update your identity independently, including pointing it to a new PDS if needed.
 
-## Recovery Keys
+## Rotation Keys
 
-Each DID document publishes two public keys: a signing key and a recovery key.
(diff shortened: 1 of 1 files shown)


axfelix:
Thanks for this! Agreed we were mixing terminology here.

PR #706, opened 2026-07-02
Fix typo in security concerns section

GerardPolloRebozado:
Corrected a typo that repeated `from` 2 times


alejandrolaguna20:
I believe that it should be something like:
>"...especially when the network locators for those hosts come from untrusted input."

As the current sentence feels a bit awkward to me.

PR #707, opened 2026-07-05
Update Japanese guides

yamarten:
Followed to the latest version of the English guide from #557, and several minor typo corrections have been added.

The reason why there are large changes in the auth guides is because the translation that has been [replaced](https://github.com/bluesky-social/atproto-website/commit/bb2d44c2d5272f0ac68f0b322e078cdedee5e3b2) once is rewinded. I'm not going to stick to the original translation, but this method was more convenient because of the consistency of the style and the translation flow.


axfelix:
Thanks!

PR #708, opened 2026-07-07
feat: More details on Account Migrations

fatfingers23:
I've added a bit more details on the Account Migration guide. Mostly just little detailed things, and added a bit more detail on others. They are all things I find myself repeating when helping orgs create migration tools. The goal was to try to put down every I knew that may not be covered somewhere to help folks who are building migration tools. Some of the details are pretty specific to that use case, or to the ref PDS. 

Happy if some of this also should homed somewhere else,  I was wanting to PR this to atproto.com so hopefully it gets the most eyes on it when building out a migration tool


axfelix:
This is a great addition! This page is slightly more complex than I want it to be right now and this PR doesn't exactly limit that complexity, but I'm basically satisfied with the TL;DR at the top and it's good to have this captured somewhere, especially the new section on blobs. Merging!

fatfingers23:
While I was writing it it felt like we needed a overview one, and a "So you want to build a migration tool" page.
Thank you so much for merging! Holler if there's any thing else I can help with the subject!

PR #709, opened 2026-07-08
OAuth spec: allow `localhost` redirect URIs in the localhost dev exception

### LINKED ISSUE #699: [Spec Change Request] Allow localhost for OAuth
It's a huge pain point that you have to do https://127.0.0.1:3000 instead of http://localhost:3000/ in local development for atproto projects. I've never seen a requirement like this before. The muscle memory is strong, and since both addresses usually work (until you try login), it's super easy to get confused why sessions don't work. I think enforcing this is hindering atproto adoption. Some anecdotal evidence that other devs feel the same way: https://bsky.app/profile/mk.gg/post/3mkhexyl7yk2m

As suggested by @ThisIsMissEm [here](https://bsky.app/profile/thisismissem.social/post/3mp2l3hlpnc2l), I'm filing this here as a request to change the spec and relax it from "enforced" and "recommended". In practice, IMO, I'd still want default examples etc to use `localhost`. But at the very least it would be nice if it didn't break.

gaearon:
Closes #699.

*(This is written by Claude so it needs a review from someone who understands OAuth well.)*

---

## What

Relaxes the "Localhost Client Development" exception in the OAuth spec so that `redirect_uri` values may use the hostname `localhost` in addition to the loopback IP literals `127.0.0.1` and `[::1]`, and adds `http://localhost/` to the default `redirect_uri` values. The `client_id` rules are unchanged (origin remains exactly `http://localhost`), as are all production redirect URI rules.

## Why

Requiring the app under development to be served from `http://127.0.0.1:PORT` instead of `http://localhost:PORT` is a recurring developer-experience trap (#699):

- Every mainstream dev server (Next.js, Vite, etc.) advertises `http://localhost:PORT`, so that is what developers open.
- OAuth session state (DPoP private key, PKCE verifier, `state`) is kept in origin-scoped browser storage. When the callback lands on `127.0.0.1` — a different origin — that state is invisible, and login fails in a way that is hard to diagnose. Both addresses serve the app fine until the login step, which makes the failure look like an SDK bug.
- The official [web OAuth tutorial](https://atproto.com/guides/oauth-tutorial) currently has to document a workaround for exactly this ("This avoids cookie issues that arise from `localhost` vs `127.0.0.1` mismatches").

## Security rationale

The loopback-literal preference comes from [RFC 8252 §8.3](https://datatracker.ietf.org/doc/html/rfc8252#section-8.3), which is guidance (NOT RECOMMENDED, not prohibited) aimed at native apps binding a local port listener, out of concern that `localhost` could resolve to a non-loopback address on misconfigured systems. That concern does not carry marginal weight in this flow:

- The application under development is already served from and browsed at the `localhost` origin, so the OAuth callback depends on `localhost` name resolution no more than the application itself does. The IP-literal rule only relocates the callback; it cannot protect an app the developer is browsing at `localhost`. This argument requires no assumptions about browser behavior.
- Browsers that treat `http://localhost` as a [secure context](https://w3c.github.io/webappsec-secure-contexts/#localhost) do so on the condition that `localhost` names are guaranteed to resolve to the loopback interface (the Secure Contexts spec makes trust conditional on exactly this guarantee; current Chrome, Firefox ≥84, and Safari all provide it).
- The interception threat RFC 8252 worries about is neutralized by this profile: PAR binds both the PKCE `code_verifier` and the DPoP key to the session at flow start, and the token endpoint requires both — so an intercepted authorization code cannot be redeemed (verified against the reference `@atproto/oauth-provider` implementation, which rejects token requests whose DPoP key does not match the JKT captured at PAR). An attacker who can exfiltrate those secrets from the app has already won regardless of redirect hostname.
- The other RFC 8252 concern — accidentally listening on non-loopback interfaces — is a property of what the dev server binds to, not of the hostname in the redirect URI, so the current rule does not mitigate it.
- This exception is development-only by construction, and remains optional for Authorization Servers.

Most major OAuth providers (GitHub, Google, Auth0, Okta) accept `http://localhost` redirect URIs for development.

## Follow-ups (out of scope here)

- Reference implementation (`@atproto/oauth-provider`) would need to accept `localhost` redirect URIs for the exception.
- Once implementations accept it, the tutorials/guides can switch their examples back to `localhost` and drop the workaround notes.
- Translations (`ko.mdx`) left to the usual translation update process.


ApexDevelopment:
+1, would make my life a lot easier.

dmackinn:
I would personally love to see this adopted, I think we all complain about it but nobody has actually tackled fixing this.

gaearon:
I didn't want to argue about merits of LLM-generated PRs tonight and kinda rageclosed this but I'll probably reopen some version of this in the future.

axfelix:
Hey! I'd like to reopen this as a proposal (like https://github.com/bluesky-social/atproto/discussions/4708).

@gaearon, are you OK with me reproducing your points here with a paraphrase, or would you like to own it? I'm happy either way. Any other context we should have for this?

gaearon:
Sure, please take it over! You're welcome to reproduce any parts, I just haven't verified them myself because I'm not a security person.

axfelix:
Great! It lives in https://github.com/bluesky-social/atproto/discussions/5255 now.

PR #741, opened 2026-08-10
Replace reference to deprecated React.FormEvent in Next.js OAuth guide

matchai:
As per `@types/react`, [“FormEvent doesn't actually exist”](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e1f7906509444c6e70ff359060af2ae034dbd369/types/react/index.d.ts#L2087).
It feels, both, more accurate and correct to use the `SubmitEvent` type for the form submission.

This PR updates the type used in the form submission code snippet.


axfelix:
Good suggestion, thank you!

PR #743, opened 2026-08-11
Fix local setup in Next.js OAuth tutorial

matchai:
Hey folks! Two quick fixes, having just gone through the guide myself:

**1. Create Next.js app with pnpm**

The tutorial has the user create the project with `npx create-next-app`, which runs the npm installation and creates an npm lockfile. The later `pnpm add` causes a reinstall via pnpm, generating a pnpm lockfile.

This PR updates the instructions to use `pnpm create next-app`, directly installing with pnpm.

**2. Add `127.0.0.1` as a development origin**

The OAuth tutorial directs readers to use `127.0.0.1`, but current Next.js versions block development resources from non-localhost origins unless:

1. The hostname is explicitly set to `127.0.0.1`, like so: `next dev --hostname 127.0.0.1`
2. A non-localhost hostname is explicitly allowlisted.

I opted for number 2 as simpler, since we're already modifying the next config in the guide.

**3. Explicitly call out that the dev script needs updating**

I missed this change since I saw the comment saying `// Add this line:`. I thought nothing else had changed in the scripts.


axfelix:
These are very good flow-polishing changes, thank you!

PR #744, opened 2026-08-13
Fix a few small hiccups in the social app tutorial

matchai:
Hey folks! A few more quick docs fixes to the social app tutorial:

**1. Specify the version of the tap installation**

As per the output, when running `go install github.com/bluesky-social/indigo/cmd/tap`:
> go: 'go install' requires a version when current directory is not in a module
        Try 'go install github.com/bluesky-social/indigo/cmd/tap@latest' to install the latest version

**2. Fix migration**

The `down` for migration `002` didn't correctly reverse the described `up`.

**3. Fix transaction committing before it finishes being used**

Since we didn't await `setCurrStatus`, an unhandled exception was thrown when inserting a new status.


axfelix:
This is very much appreciated! Had to verify why those missing `await`s weren't causing a problem for me previously but it makes sense to add them now.

PR #765, opened 2026-09-04
Fix code in lexicon guide

yamarten:
Fix the sample code inconsistency in the [lexicon guide](https://atproto.com/guides/lexicon).

* the NSID used in the SDK call example
* [`key` field](https://atproto.com/specs/lexicon#record) missing in lexicon sample


axfelix:
This is a fine correction, thanks!

